Issue #2, opened 2021-09-16
Extract hexagon icon borders from each coloured card


micahstairs:
For each of the 5 colors, we need to extract the hexagon icon border, and place it in a PNG with a translucent background. 

Here are the source images:
- [ ] [blue_source.png](../tree/main/misc/card_icon_borders/blue_source.png)
- [ ] [green_source.png](../tree/main/misc/card_icon_borders/green_source.png)
- [ ] [purple_source.png](../tree/main/misc/card_icon_borders/purple_source.png)
- [ ] [red_source.png](../tree/main/misc/card_icon_borders/red_source.png)
- [ ] [yellow_source.png](../tree/main/misc/card_icon_borders/yellow_source.png)

Please remove the card-specific icon from inside the hexagon and just fill it white. So, there should be 3 components to each image you create: the outer part should be translucent, then inside that there should be a colored hexagon border, and then inside that it should be a solid white.

Please place each hexagon icon border in a separate image (naming it something like `blue_hexagon_icon_border`).

**NOTE:** Make sure not to resize the image or lower its quality in any way.

PR #7, opened 2021-09-17
hexagon icon borders

normaljosh:
Hexagon icon borders, see #2


micahstairs (changes requested):
Looks pretty good! Can you crop each of the images so that the icons are centered, with only a few pixels around the outside of the icons?

normaljosh:
Sure thing, done!

micahstairs (approved):
Thanks!

Issue #6, opened 2021-09-17
Extract shape (under age number) from each coloured card 


micahstairs:
For each of the 5 colors, we need to extract the shape used underneath the age number (there are different shapes to help those who are colorblind).

Here are the source images:
- [ ] [blue_source.png](../tree/main/misc/card_icon_borders/blue_source.png)
- [ ] [green_source.png](../tree/main/misc/card_icon_borders/green_source.png)
- [ ] [purple_source.png](../tree/main/misc/card_icon_borders/purple_source.png)
- [ ] [red_source.png](../tree/main/misc/card_icon_borders/red_source.png)
- [ ] [yellow_source.png](../tree/main/misc/card_icon_borders/yellow_source.png)

Please remove the number from inside the shape. Use a translucent background for the outside of the shape.

Please place each shape in a separate image (naming it something like `blue_shape`).

**NOTE:** Make sure not to resize the image or lower its quality in any way. Also make sure all 5 images are the same size, with the icon centered inside of it.

PR #8, opened 2021-09-18
Upload files for shapes for the age numbers

jacksonmead:
Attempting to upload shape files for age numbers for #6


micahstairs (changes requested):
Looks pretty good!

The only thing is that I think the background is still white, not transparent! Are you able to fix that? If not, I can accept this pull request and then we can find someone else to touch up the transparency.

micahstairs:
Never mind, I think the background is already transparent. Thanks so much for taking the time to do this!

micahstairs (approved):

jacksonmead:
OK, whew, I was worried after that first comment :-) Thanks so much for taking on maintenance of Innovation!

Issue #13, opened 2021-09-19
Extract 6 square icons from cards


micahstairs:
We need to extract each of the 6 square icons, and place them in a PNG. 

Here are the source images:
- [ ] [castle_and_crown_source.png](../tree/main/misc/square_icons/castle_and_crown_source.png)
- [ ] [factory_and_clock_source.png](../tree/main/misc/square_icons/factory_and_clock_source.png)
- [ ] [leaf_and_lightbulb_source.png](../tree/main/misc/square_icons/leaf_and_lightbulb_source.png)

Please place each in a separate image (naming it something like `castle.png`). Make sure not to resize the image or lower its quality in any way.

Please make sure all icons are the exact same size.

**NOTE:** Before you begin working on this, please **assign this issue to yourself**. If you do not do this, then there's the possibility that 2 people will end up doing the same task, which is not a good use of time.

jacksonmead:
@micahstairs It looks like factory_and_clock_source.png and leaf_and_lightbulb_source.png are 404.

micahstairs:
@jacksonmead sorry about that, fixed the links!

PR #14, opened 2021-09-20
Upload 6 square icons

jacksonmead:
Icons for #13 . These were slightly different sizes in the source files (because of anti-aliasing, I assume). I made them all 137 pixels square.


micahstairs (approved):
Thanks for making the sizes consistent!

Issue #21, opened 2021-09-24
Build ImageMagick commands to extract hexagon icons from cards


micahstairs:
For a given card, we need to extract the icon and place it inside the corresponding `*_hexagon_icon_border.png` template. The border templates are found in https://github.com/micahstairs/bga-innovation/tree/main/misc/card_icon_borders.

See https://github.com/micahstairs/bga-innovation/tree/main/misc/cards for all of the source images.

There are 4 possible locations for the hexagon on the card, so this task will involve creating 4 InageMagick commands (one to extract the icon from each location).

Here's a discussion about how to do something similar: https://legacy.imagemagick.org/discourse-server/viewtopic.php?t=35510&p=163645.

Once we have these commands written, I'll file separate issues to write the commands for all the cards. For now we just want to get sample commands working.

normaljosh:
Might be possible to do as a one-liner but I was running into some issues with that, so it's one command to crop out the black hexagon, and another one to paste onto the border. 

commands are in the file imagemagick_commands/hexagon_icons.txt, sample hexagons are in the hexagon_icons folder.

micahstairs:
Great work! A 2-liner is just as useful as a 1-liner here!

PR #23, opened 2021-09-25
txt file with imagemagick commands at four positions

normaljosh:
Imagemagick commands for issue #21 . Commands for all four positions are in a .txt file at imagemagick_commands/hexagon_icons.txt, example images are in the hexagon_icons folder.


micahstairs (approved):
This is amazing! Thanks for taking the time to figure out this command.

This will be so helpful in extracting hundreds of icons from the various expansions.

Issue #32, opened 2021-09-28
Make spritesheet with transparent background


micahstairs:
We need to take https://github.com/micahstairs/bga-innovation/blob/main/misc/hexagon_icons/base_hexagon_spritesheet.png a couple steps further:

- By making the background **transparent**.
- We also need to put a few more icons in some of the open space. At the end of the 2nd row, let's put https://github.com/micahstairs/bga-innovation/blob/main/misc/card_shapes/colorblind_indicator_spritesheet.png. At the end of the 3rd row, let's put the 5 achievement icons from https://github.com/micahstairs/bga-innovation/tree/main/misc/special_achievement_icons. And then at the end of the 4th row, let's put the 5 **rotated** achievement icons from.

You **must** use ImageMagick for this task!

**NOTE:** Before you begin working on this, please **assign this issue to yourself**. If you do not do this, then there's the possibility that 2 people will end up doing the same task, which is not a good use of time.

PR #36, opened 2021-09-29
transparency and added icons for base shape spritesheet

normaljosh:
Added icons and changed to transparent background for base_hexagon_spritesheet as per issue #32


normaljosh:
OK, resized to 120x120 if not already smaller than that, and snapped to the same grid!

micahstairs (approved):
Amazing! This is much easier to use, thanks.

Issue #40, opened 2021-10-01
Swap location of green and red shape icons


micahstairs:
We need to switch the location of the green and red shape icons in the spritesheet. In the implementation of the game, we need to use the same order of the colors everywhere or it will be a little confusing. So we want: blue, red, green, yellow, purple.

**NOTE:** Before you begin working on this, please **assign this issue to yourself**. If you do not do this, then there's the possibility that 2 people will end up doing the same task, which is not a good use of time.

PR #42, opened 2021-10-06
reorder colorblind indicators, trim whitespace in spritesheet

normaljosh:
See issue #40 

- Reordered colorblind indicators as requested 
- Trimmed whitespace when making spritesheet, so icons should resize to same size
- Cleaned up green icon (there were some almost-invisible pixels in what should be the transparent area)


micahstairs:
Can you make the square, hexagon, and circle a bit smaller? I know the widths/heights are the same on all of the images though, but these one appear larger due to their surface areas. I think shrinking these 3 shapes a bit will make them seem more balanced.

normaljosh:
Here's my go at it, though now they're not exactly aligned the same. Does that still work? It was more complicated than you'd think

micahstairs (approved):
This looks great! Thanks so much.

With CSS, it's way cleaner for me to offset the location of the images a bit than it is to scale each icon differently.

Issue #82, opened 2021-11-01
Implement Artifact Age 1 Ark of the Covenant (ID 123)


ultimatefiend:
**NOTE:** Before you begin working on this, please **assign this issue to yourself**. If you do not do this, then there's the possibility that 2 people will end up doing the same task, which is not a good use of time.

ultimatefiend:
Detecting whether a top card is an artifact card might be a problem.  I think for now I can use the ID of the card to determine this.  It would be nice if the Database contained this information.

micahstairs:
Using the ID is a reasonable workaround for now. Be sure to leave a TODO for me!

PR #84, opened 2021-11-02
Ark of the Covenant

ultimatefiend:
#82 - Using the ID to detect whether its an artifact card or not.  It would be better if that came from the database.

### DIFF
--- a/innovation.game.php
+++ b/innovation.game.php
@@ -7375,6 +7375,11 @@ class Innovation extends Table
                 $step_max = 2; // --> 2 interactions: see B
                 break;
 
+            // id 123, Artifacts age 1: Ark of the Covenant
+            case "123N1":
+                $step_max = 1; // --> 1 interactions: see B
+                break;
+				
             // id 124, Artifacts age 1: Tale of the Shipwrecked Sailor
             case "124N1":
                 $step_max = 2; // --> 2 interactions: see B
@@ -9727,7 +9732,21 @@ class Innovation extends Table
                 'score_keyword' => true
             );
             break;
-        
+			
+        // id 123, Artifacts age 1: Ark of the Covenant
+        case "123N1A":
+            // "Return a card from your hand."
+            $options = array(
+                'player_id' => $player_id,
+                'can_pass' => false,
+                'n' => 1,
+                'owner_from' => $player_id,
+                'location_from' => 'hand',
+                'owner_to' => 0,
+                'location_to' => 'deck',
+            );
+            break;
+			
         // id 124, Artifacts age 1: Tale of the Shipwrecked Sailor
         case "124N1A":
             // "Choose a color"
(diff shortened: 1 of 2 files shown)


innovation.game.php
```diff
@@ -10486,6 +10505,58 @@ function stInterInteractionStep() {
...
+                case "123N1A":
+					$players = self::loadPlayersBasicInfos();
+					
+					if ($n > 0)
+					{ // Unsaid rule: the player must have at least one card to show from his hand, else, the effect can't continue
+                        $color = self::getGameStateValue('color_last_selected');
+                            
+                        foreach($players as $all_player_id => $player) 
+						{
+                            // TODO : is there a better way to get all top cards?
```
micahstairs, line 10518: Can you write a new `getTopCardsOnBoard` function? It will be quite similar to `getTopCardOnBoard` except we we drop the bits which restrict the query to a specific color, and you'll also need to use `getCollectionFromDb` instead of `getObjectFromDB` since we are grabbing an array of elements instead of a single element.

I just ran the following query on the database and it works:

```
                SELECT
                    *
                FROM
                    card
                WHERE
                    card.owner = {player_id} AND
                    card.location = 'board' AND
                    card.position = (
                        SELECT
                            MAX(position) AS position
                        FROM
                            card
                        WHERE
                            owner = {player_id} AND
                            location = 'board' AND
                            color = card.color
                    )
```

innovation.game.php
```diff
@@ -10486,6 +10505,58 @@ function stInterInteractionStep() {
...
+								($top_purple_card == null || $top_purple_card['id'] < 110))
+							{
+								// No artifacts as a top card, transfer all cards to the original player's
+								// scorepile
+								$top_card = self::getTopCardOnBoard($all_player_id, $color);
+								while ($top_card !== null)
+								{	
+									// "Transfer all cards of the same color from the boards of all players
+									// with no top artifacts to your score pile."
+									// This is a "transfer" so monument cannot be achieved with this dogma.
```
micahstairs, line 10539: Let's just omit this comment to keep things shorter.

innovation.game.php
```diff
@@ -10486,6 +10505,58 @@ function stInterInteractionStep() {
                     }
                     break;
                 
+				// id 123, Artifacts age 1: Ark of the covenant
+                case "123N1A":
+					$players = self::loadPlayersBasicInfos();
+					
+					if ($n > 0)
+					{ // Unsaid rule: the player must have at least one card to show from his hand, else, the effect can't continue
+                        $color = self::getGameStateValue('color_last_selected');
```
micahstairs, line 10514: Can you rename this variable to "return_color" for additional clarity? This will also free up the color variable in case we need to use it in a for loop.

innovation.game.php
```diff
@@ -10486,6 +10505,58 @@ function stInterInteractionStep() {
...
+							if (($top_blue_card == null || $top_blue_card['id'] < 110) &&
+								($top_red_card == null || $top_red_card['id'] < 110) &&
+								($top_green_card == null || $top_green_card['id'] < 110) &&
+								($top_yellow_card == null || $top_yellow_card['id'] < 110) &&
+								($top_purple_card == null || $top_purple_card['id'] < 110))
+							{
+								// No artifacts as a top card, transfer all cards to the original player's
+								// scorepile
+								$top_card = self::getTopCardOnBoard($all_player_id, $color);
+								while ($top_card !== null)
```
micahstairs, line 10535: The structure of the while loop can be simplified to:
```
while (($top_card = self::getTopCardOnBoard($all_player_id, $color)) !== null) {	
  self::transferCardFromTo($top_card, $player_id, 'score');
}
```

I believe that should work in PHP? I know it does in other languages.

micahstairs: Definitely still leave the comments in here quoting the card.

innovation.game.php
```diff
@@ -10486,6 +10505,58 @@ function stInterInteractionStep() {
...
+									// with no top artifacts to your score pile."
+									// This is a "transfer" so monument cannot be achieved with this dogma.
+									self::transferCardFromTo($top_card, $player_id, 'score');
+									
+									$top_card = self::getTopCardOnBoard($all_player_id, $color);
+								}
+                            }
+                        }
+                    }
+					// "If Ark of the Covenant is a top card on any board, transfer it to your hand."
```
micahstairs, line 10547: It seems there are multiple cards which will require us to check if a specific card is the top card on a board.

Can you write a `getIfTopCard` function which takes a card ID and returns the card only if it a top card on any board? I've tested the following SQL query:

```
                SELECT
                    *
                FROM
                    card
                WHERE
                    card.id = {id} AND
                    card.location = 'board' AND
                    card.position = (
                        SELECT
                            MAX(position) AS position
                        FROM
                            card
                        WHERE
                            owner = card.owner AND
                            location = 'board' AND
                            color = card.color
                    )
```

This will drastically simplify this code.

innovation.game.php
```diff
@@ -10486,6 +10505,58 @@ function stInterInteractionStep() {
...
+						{
+                            // TODO : is there a better way to get all top cards?
+							$top_blue_card = self::getTopCardOnBoard($all_player_id, 0);
+							$top_red_card = self::getTopCardOnBoard($all_player_id, 1);
+							$top_green_card = self::getTopCardOnBoard($all_player_id, 2);
+							$top_yellow_card = self::getTopCardOnBoard($all_player_id, 3);
+							$top_purple_card = self::getTopCardOnBoard($all_player_id, 4);
+							
+							// TODO : if there is an artifact card on top of any pile, then prevent the rest of the dogma.
+							if (($top_blue_card == null || $top_blue_card['id'] < 110) &&
```
micahstairs, line 10526: I suppose this will just turn into a for loop over the returned array.

micahstairs (changes requested):
Thanks for tackling this card! Here's the first round of my comments. I made a few suggestions on how to simplify this code, and how to make it more performant (by making less SQL queries).

innovation.game.php
ultimatefiend: I'm not sure you want to trust me with SQL...heh.  I can give it a try.

micahstairs: Haha I already wrote out the tested SQL query so it's just a matter of putting it in a PHP function!

innovation.game.php
ultimatefiend: That wouldn't work in C that's for sure.  I can try it in PHP and see what happens.

innovation.game.php
ultimatefiend: Yes.

innovation.game.php
ultimatefiend: sounds good

innovation.game.php
ultimatefiend: Yes

micahstairs (approved):
Thanks!

PR #86, opened 2021-11-02
Implement "Chronicle of Zuo" card

ultimatefiend:
Part of #88.

### DIFF
--- a/innovation.game.php
+++ b/innovation.game.php
@@ -7309,9 +7309,9 @@ class Innovation extends Table
                 self::executeDraw($player_id, 2, 'hand');
                 break;
 
-			// id 114, Artifacts age 1: Papyrus of Ani
+            // id 114, Artifacts age 1: Papyrus of Ani
             case "114N1":
-				$step_max = 1; // --> 1 interactions: see B
+                $step_max = 1; // --> 1 interactions: see B
                 break;
 
             // id 115, Artifacts age 1: Pavlovian Tusk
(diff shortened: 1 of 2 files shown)


innovation.game.php
```diff
@@ -7389,6 +7389,49 @@ function stPlayerInvolvedTurn() {
             case "124N1":
                 $step_max = 2; // --> 2 interactions: see B
                 break;
+
+            // id 127, Artifacts age 2: Chronicle of Zuo
+            case "127N1":
+                $players = self::loadPlayersBasicInfos();
+				$min_towers = 100;
```
micahstairs, line 7396: I'm not a fan of the fact that this default value is so low, but my suggestion below will eliminate this anyway!

micahstairs: Instead of using a for loop to iterate over the players, lets inline a function here which grabs the min value for us.

```$min_towers = self::getUniqueValueFromDB(self::format("SELECT MIN(player_icon_count_4) FROM player WHERE player_id != {player_id}", array('player_id' => $player_id))```

Given that this query is highly specific and pretty short, I don't think there's much value in refactoring it out into a separate function. Especially since this would be hard to give a concise name.

**Review by micahstairs — CHANGES_REQUESTED** [review]

File: innovation.game.php [hunk 2]
```diff
@@ -7389,6 +7389,49 @@ function stPlayerInvolvedTurn() {
...
+                }
+				if ($this_player_icon_counts[4] < $min_towers)
+				{
+					$card = self::executeDraw($player_id, 2, 'hand'); // "If you have the fewest towers, draw a 2"
+				}
+				if ($this_player_icon_counts[1] < $min_crowns)
+				{
+					$card = self::executeDraw($player_id, 3, 'hand'); // "If you have the fewest crowns, draw a 3"
+				}
+				if ($this_player_icon_counts[3] < $min_bulbs)
```
micahstairs, line 7430: What's the expected behavior for this card if the player is tied for the fewest with another player?

ultimatefiend: Great question.  I was assuming that it had to be strictly less than all other players to get the draw.  If it doesn't, then the query you have earlier will work.

micahstairs: It's probably safe to assume for now that it's strictly less unless we hear otherwise from Chris.

ultimatefiend:
new content uploaded with the sql search.

ultimatefiend:
I don't think that simplifies the logic.  It adds an extra line and the existing line is longer when adding the parameterization.

PR #91, opened 2021-11-03
Transcribe age 3 artifact cards

jacksonmead:
Artifacts Age 3 card transcriptions.


micahstairs:
Thanks for putting this together! Just a few minor suggestions for next time:

1) Please make sure your branch is based on `main-artifacts` instead of `main`. When I switch the branch it pulled in extra commits, so I had to commit your changes directly myself (I'm going to close this pull request for this reason)
2) There's now a new `'i_demand_effect_1_is_compel' => false,` line which needs to go after the `i_demand_effect_1` line. We set this variable to true only if there is a compel, otherwise it will be false.
3) Please order the cards: blue, red, green, yellow, purple (within each color we can do the cards in the order that they appear in the source files we are working with). Now that we have this order established, you'll also be able to put in the card IDs next time.

Thanks! I'll be sure to include these notes in the next issue I file for card transcriptions.

PR #110, opened 2021-11-10
Age 3 artifacts (6 of them)

ultimatefiend:
Philosopher's, beauvais, along the river, dunhuang, excalibur, and mjolnir.


micahstairs:
I made a few simplifications to the code.

However, I noticed that `Dunhuang Star Chart` isn't quite working yet. Please test it when there is more than one card in your hand. I expect we will need to move the draw to a second interaction (even though we don't need an explicit action from the player).

PR #111, opened 2021-11-11
Normaljosh necronomicon

normaljosh:
Completed except for the green case (unsplay all piles), I'm a bit stuck here. 

Going to post in the discord as well, but my current method gets the error 

```
Server syntax error:
Notice: Undefined variable: location_from in /var/tournoi/release/games/innovationnormaljosh/999999-9999/innovationnormaljosh.game.php on line 5627
{"status":1,"data":{"valid":1,"data":null}}
```

I'm thinking something's going wonky with the code around lines 4003, where $rewritten options isn't setting a value for location_from, but to be honest I'm not entirely clear on what's happening around that code. It seems like the "else" should be triggered since I have a splay_direction key in my options array, but for some reason it isn't. 

Green case that isn't working: 
``` php
            // "If green, unsplay all piles"
            else if (self::getGameStateValue('auxiliary_value') == 2 /* green */) {
                self::notifyGeneralInfo(clienttranslate("This card is green."));
                $splayed_dcolors = array();
                for($color=0; $color<5; $color++) {
                    if (self::getCurrentSplayDirection($player_id, $color)!==0 /* splayed */) {
                        $splayed_colors[] = $color;
                    }
                }
                $options = array(
                    'player_id' => $player_id,
                    'n' => sizeof($splayed_colors),
                    'can_pass' => false,
                    
                    'splay_direction' => 0 /* unsplay */,
                    'color' => $splayed_colors
                );
            };
```
else that isnt being triggered: 
``` php
        if (!array_key_exists('splay_direction', $rewritten_options)) {
             $rewritten_options['splay_direction'] = -1;
        }
        else { // This is a choice for splay
            $rewritten_options['owner_from'] = $player_id;
            $rewritten_options['location_from'] = 'board'; // Splaying is equivalent as selecting a board card, by design
            $number_of_cards_on_board = self::countCardsInLocation($player_id, 'board', null, false, true);
            $splay_direction = $rewritten_options['splay_direction'];
            $colors = array();
```

### DIFF
--- a/dbmodel.sql
+++ b/dbmodel.sql
@@ -251,3 +251,4 @@ INSERT INTO `card` (`id`, `type`, `age`, `color`, `spot_1`, `spot_2`, `spot_3`, 
 (128, 1, 2, 1, 0, 4, 4, 4, 4, 0, 'deck', 0, NULL, FALSE),
 (131, 1, 2, 3, 2, 0, 1, 2, 2, 0, 'deck', 0, NULL, FALSE),
 (132, 1, 2, 3, 1, 1, 1, 0, 1, 0, 'deck', 0, NULL, FALSE);
+(143, 1, 3, 4, 0, 2, 2, 2, 2, 0, 'deck', 0, NULL, FALSE);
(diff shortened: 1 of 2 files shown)


normaljosh:
OK, moved it out of the options array and it's working now, thanks for the advice! It should be ready to merge in.

micahstairs (approved):
Thanks, this is great! I made a few simplifications myself, but well done!

PR #131, opened 2021-11-19
Kahlia age5 : boerhavve, principia, hudson bay, barque

ultimatefiend:


innovation.game.php
```diff
@@ -7979,6 +7979,36 @@ function stPlayerInvolvedTurn() {
...
+            case "156N1":
+                $step_max = 1;
+                break;
+            
+            // id 159, Artifacts age 5: Barque-Longue La Belle
+            case "159N1":
+                do {
+                    $card = self::executeDraw($player_id, 5, 'board'); // "Draw and meld a 5"
+                    // If the drawn card is not green, repeat this effect.
+                } while($card['color'] == 0 || $card['color'] == 1 || $card['color'] == 3 || $card['color'] == 4);
```
micahstairs, line 7998: How about just checking `$card['color'] != 2`?

innovation.game.php
```diff
@@ -10878,7 +10908,86 @@ function stInteractionStep() {
...
+        // id 156, Artifacts age 5: Principia
+        case "156N1A":
+            $ages_on_top = array();
+            for($color = 1; $color < 5; $color++) {
+                $topcard = self::getTopCardOnBoard($player_id, $color);
+                if ($topcard !== null){ // skip empty piles
+                    $ages_on_top[] = $topcard['age']; // Record all non-blue ages
+                }
+            }
+            self::setGameStateValueFromArray('auxiliary_value', $ages_on_top);
```
micahstairs, line 10963: The array you pass into setGameStateValueFromArray cannot have duplicate values or it won't work correctly.

I haven't tested the following code, but I believe something like this would work (other cards do something similar):

$ages_on_top = array();
self::setGameStateValueAsArray('auxiliary_value', $ages_on_top);
for($color = 1; $color < 5; $color++) { // non-blue
    $top_card = self::getTopCardOnBoard($player_id, $color);
    if ($top_card !== null) {
        $ages_on_top[] = $top_card['age'];
        if (!in_array($top_card['age'], $ages_on_top)) {
            $ages_on_top[] = $top_card['age'];
            self::setGameStateValueFromArray('auxiliary_value', $ages_on_top);
        }
    }
}

innovation.game.php
```diff
@@ -11727,8 +11836,38 @@ function stInterInteractionStep() {
...
                     }
                     break;
-                }
 
+                // id 155, Artifacts age 5: Boerhavve Silver Microscope
+                case "155N1B":
+                    // "Draw and score a card of value equal to the sum of the values of the cards returned."
+                    $first_age = self::getGameStateValue('auxiliary_value');
+                    $second_age = self::getGameStateValue('age_last_selected');
+                    self::executeDraw($player_id, $first_age + $second_age, 'score');
```
micahstairs, line 11845: Don't we only want to draw if we know that two cards were actually returned?

Maybe check that `$first_age > 0` and `$second_age > 0`?

innovation.game.php
```diff
@@ -11727,8 +11836,38 @@ function stInterInteractionStep() {
...
+                        self::executeDraw($player_id, $card_age + 1, 'board');
+                    }
+                    break;
+
+                // id 160, Artifacts age 5: Hudson's Bay Company Archives
+                case "160N1A":
+                    $color = self::getGameStateValue('color_last_selected');
+                    $board = self::getCardsInLocation($player_id, 'board', null, false, true);
+                    $pile = $board[$color];
+                    if (count($pile) > 1) {
```
micahstairs, line 11864: There's no need to check the count of the pile, since this is done inside the splay method.

micahstairs (changes requested):
Looks pretty good! I left a few comments on some things that need changed though.

innovation.game.php
ultimatefiend: I did that originally, but it it was never true.  Actually, I did !==.

innovation.game.php
ultimatefiend: no card returned would be 0.

innovation.game.php
ultimatefiend: ah ok.  its hard to know what I should check for...heh

innovation.game.php
ultimatefiend: if it said, "if you do" then it would matter if two cards are returned.

innovation.game.php
micahstairs: !== is a bit different than != since !== also requires the data to be the same type. It's possible that the type of data being pulled from the database is in a slightly format than a number like 2.

I'd recommend trying != and seeing if that works here!

innovation.game.php
micahstairs: Hmm let's think, so the card says "Draw and score a card of value equal to the sum of the values of the cards returned". I guess that means if you only return one card, then you draw and score a card of that same value? And if they return two cards then you take the sum?

Please test that both cases work.

ultimatefiend: right
I've tested both cases including an empty hand.

innovation.game.php
ultimatefiend: will do.

innovation.game.php
ultimatefiend: Yup, that works!  Thanks.

innovation.game.php
ultimatefiend: hmm...this array needs to have duplicates in it.  (i.e. you can return more than one card of the same age)

This limitation of arrays is really annoying...heh.

ultimatefiend: After looking at the code for the array encoding, I have an idea for a different type of encoding which will make arrays of "ages" much easier.  The maximum value of the integers in this framework is insanely small though (0x8000000).  Why not a full 32 bit number?  I think adding additional "auxiliary_value" fields will be needed moving forward to augment the size of the arrays.

ultimatefiend: Ok, this is what I have:

   // This function encodes multiple integers into a single array
    // which can be decoded by getGameStateBase16Array.
    // Note that due to the maximum data value of 0x8000000 that only
    // 5 elements can be encoded in this manner.
    // This function is ideal for encoding multiple age values into an array.
    function setGameStateBase16Array($key, $array) {
        if (count($array) > 5) {
            // Too many elements to encode.  Throw an error.
        }
        $encoded_value = 0;
        foreach($array as $value) {
            $encoded_value = $encoded_value * 16 + $value;
        }
        $encoded_value = $encoded_value * 6 + count($array);
        self::setGameStateValue($key, $encoded_value);
    }
    
    // This is the decoding partner of setGameStateBase16Array
    function getGameStateBase16Array($key) {
        $return_array = array();
        $encoded_value = self::getGameStateValue($key);
        $count = ($encoded_value % 6); // count is first
        $encoded_value /= 6;
        for ($i = 0; $i < $count; $i++) {
            $return_array[] = $encoded_value % 16;
            $encoded_value /= 16;
        }
        return $return_array;
    }
    
It can have a maximum size of 5 and a minimum of 0.  I tested it out with Principia and it works pretty well.

micahstairs: Great! Let me know when you've pushed your changes and want me to take a look at this PR.

micahstairs (approved):
Looks great! I only made a few minor changes.

PR #139, opened 2021-11-23
Kahlia age7 - corvette, jedlik, singer

ultimatefiend:
Added a new function that counts visible cards.
Otherwise, the cards are pretty straightforward


micahstairs (approved):
Solid work here!

PR #140, opened 2021-11-24
Kahlia age5 - bill, almira, cricket

ultimatefiend:


innovation.game.php
```diff
@@ -2001,6 +2001,11 @@ function getTransferInfoWithTwoPlayersInvolved($location_from, $location_to, $pl
                 $message_for_others = clienttranslate('{player must} transfer {number} top {card} from his board to {opponent_name}\'s achievements');
                 break;
     
+            case 'board->none':
+                $message_for_player = clienttranslate('{You must} select a pile from {your} board');
+                $message_for_opponent = clienttranslate('{player must} select a pile from {your} board');
+                $message_for_others = clienttranslate('{player must} select a pile from his board');
+                break;
```
micahstairs, line 2008: This can be removed (see below).

innovation.game.php
```diff
@@ -11164,6 +11195,29 @@ function stInteractionStep() {
...
+            $options = array(
+                'player_id' => $player_id,
+                'n' => 1,
+                'can_pass' => false,
+                'color' => $more_vis_cards,
+                
+                'owner_from' => $player_id,
+                'location_from' => 'board',
+                'owner_to' => $launcher_id,
+                'location_to' => 'none'
```
micahstairs, line 11217: We can simplify the implementation this card by using `'choose_color' => true`. For example, see "Mask of Warka".

micahstairs: If you follow "Mask of Warka" closely, you can figure out where the other code for this card belongs.

micahstairs (changes requested):
Great work! I described a change which makes choosing a color match how we implement it for other cards.

innovation.game.php
micahstairs: Never mind, the suggestion won't work.

innovation.game.php
micahstairs: Never mind, the suggestion won't work, since that doesn't support choosing a subset of colors, and it looks non-trivial to extend support for that.

innovation.game.php
```diff
@@ -2003,6 +2003,11 @@ function getTransferInfoWithTwoPlayersInvolved($location_from, $location_to, $pl
                 $message_for_others = clienttranslate('{player must} transfer {number} top {card} from his board to {opponent_name}\'s achievements');
                 break;
     
+            case 'board->none':
```
micahstairs: I removed this "board->none" case since we are no longer using it.

innovation.game.php
```diff
@@ -5514,6 +5519,8 @@ function argSelectionMove() {
 
             // id 122, Artifacts age 1: Mask of Warka
             case "122N1A":
+            // id 157, Artifacts age 5: Bill of Rights
```
micahstairs: In order to be consistent with the rest of our code (and to keep the cards in order), I split this into two separate cases, even though it results in a little bit of extra duplicated code.

micahstairs (approved):
Looks great, thanks!

Issue #88, opened 2021-11-02
Implement Age 2 cards for Artifacts Expansion


micahstairs:
This issue will track the implementation of all Age 2 cards for the Artifacts expansion. Please update this comment before working on any of these cards (if the work is unblocked) so that we don't have duplicate work being done. We no longer need to file a separate issue for each card, since a consolidated list of the cards will keep us more organized.

  - [x] **125 - Seikilos Epitaph**
  - [x] **126 - Rosetta Stone**
  - [x] **127 - Chronicle of Zuo**
  - [x] **128 - Babylonian Chronicles**
  - [x] **129 - Holy Lance**
  - [x] **120 - Baghdad Battery**
  - [x] **131 - Holy Grail**
  - [x] **132 - Terracotta Army**
  - [x] **133 - Dead Sea Scrolls**
  - [x] **134 - Cyrus Cylinder**

PR #141, opened 2021-11-24
Kahlia age2 1 - holy lance 

ultimatefiend:
#88


innovation.game.php
```diff
@@ -7855,12 +7855,30 @@ function stPlayerInvolvedTurn() {
...
                 break;
+
+            // id 129, Artifacts age 2: Holy Lance
+            case "129C1":
+                $step_max = 1; // --> 1 interaction: see B
+                break;
+
+            case "129N1":
+                // If Holy Grail is a top card on your board, you win.
+                $top_yellow_card = self::getTopCardOnBoard($player_id, 3);
```
micahstairs: NOTE: I simplified this using the existing `getIfTopCardOnBoard` function.

innovation.game.php
```diff
@@ -10651,6 +10669,32 @@ function stInteractionStep() {
...
+        case "129C1A":
+            // Find colors with top artifacts
+            $artf_colors = array();
+            for ($color = 0; $color < 5; $color++)
+            {
+                $top_card = self::getTopCardOnBoard($player_id, $color);
+                if ($top_card !== null && $top_card['type'] == 1) {
+                    $artf_colors[] = $color;
+                }
+            }
```
micahstairs: Instead of using a separate query to fetch the top Artifacts and needing to replicate this code when we want to do this in other places, I've added an option like `'type' => array(1)` that we can use.

micahstairs (approved):

PR #148, opened 2021-11-28
Kahlia age8 - parnell, ocean, meiji, plush

ultimatefiend:


innovation.game.php
```diff
@@ -8386,6 +8392,54 @@ function stPlayerInvolvedTurn() {
...
+                if($card['spot_2'] == 6){
+                    $count_clocks++;
+                }
+                if($card['spot_3'] == 6){
+                    $count_clocks++;
+                }
+                if($card['spot_4'] == 6){
+                    $count_clocks++;
+                }                
+                if ($count_clocks == 3) {
```
micahstairs, line 8413: Can you refactor this into a countResources($card, $icon) function? You could put it right near the existing hasRessource() function.

innovation.game.php
```diff
@@ -8386,6 +8392,54 @@ function stPlayerInvolvedTurn() {
...
+                }
+                break;
+
+            // id 189, Artifacts age 8: Ocean Liner Titanic
+            case "189N1":
+                for($color = 0; $color < 5; $color++) {
+                    $card = self::getBottomCardOnBoard($player_id, $color);
+                    if($card != null) {
+                        // Score all bottom cards from your board.
+                        $card = self::transferCardFromTo($card, $player_id, 'score', false, true);
```
micahstairs, line 8429: In order to make it more obvious what that "true" is for, let's add a comment:

$card = self::transferCardFromTo($card, $player_id, 'score', false, /*score_keyword=*/ true);

innovation.game.php
```diff
@@ -11652,6 +11706,47 @@ function stInteractionStep() {
             );
             break;            
 
+        // id 190, Artifacts age 8: Meiji-Mura Stamp Vending Machine
+        case "190N1A":
+            // Return a card from your hand.
+            $options = array(
+                'player_id' => $player_id,
+                'can_pass' => false,
+                'n' => 1,
```
micahstairs, line 11715: To be consistent with other card I always put the "n" between "player_id" and "can_pass"

innovation.game.php
```diff
@@ -12615,6 +12710,29 @@ function stInterInteractionStep() {
...
                     break;
 
+                // id 190, Artifacts age 8: Meiji-Mura Stamp Vending Machine
+                case "190N1A":
+                    // Draw and score three cards of the returned card's value.
+                    $age_to_score = self::getGameStateValue('age_last_selected');
+                    self::executeDraw($player_id, $age_to_score, 'score');
+                    self::executeDraw($player_id, $age_to_score, 'score');
+                    self::executeDraw($player_id, $age_to_score, 'score');
+                    break;
```
micahstairs, line 12720: I think you need to put this inside of a n > 0 check, otherwise it won't behave correctly if you don't have a card to return.

material.inc.php
```diff
@@ -1922,7 +1922,7 @@
 185 => array('name' => clienttranslate('Parnell Pitch Drop'),
     'i_demand_effect_1' => null,
     'i_demand_effect_1_is_compel' => false,
-    'non_demand_effect_1' => clienttranslate('Draw and meld a card of value one higher than the highest top card on your board. If the melded card has three ${icon_6}, you win.'),
+    'non_demand_effect_1' => clienttranslate('Draw and meld a card of value one higher than the highest top card on your board.  If the melded card has 3 ${icon_6}, you win.'),
```
micahstairs, line 1925: Can you please remove this extra space that you added?

micahstairs: Can you change this "3" back to "three", please?

I'm curious what you touched this file haha.

micahstairs (changes requested):
I left some comments!

innovation.game.php
ultimatefiend: I didn't think this would come up again, but I can do that.  There are no cards in echoes that do this that I know of.

innovation.game.php
ultimatefiend: sure.

innovation.game.php
ultimatefiend: sure

innovation.game.php
ultimatefiend: There isn't an "if you do", so it would result in 0s for the returned card.

material.inc.php
ultimatefiend: oh it looks like someone else fixed the text.  we can ignore the change to this file.

innovation.game.php
micahstairs: Great catch!

material.inc.php
micahstairs: Okay I reverted it.

micahstairs (approved):
Thanks, this looks great!

Issue #135, opened 2021-11-22
Implement Age 6 cards for Artifacts Expansion


micahstairs:
This issue will track the implementation of all Age 6 cards for the Artifacts expansion. Please update this comment before working on any of these cards (if the work is unblocked) so that we don't have duplicate work being done. We no longer need to file a separate issue for each card, since a consolidated list of the cards will keep us more organized.

  - [x] **165 - Kilogram of the Archives**
  - [x] **166 - Puffing Billy**
  - [x] **167 - Frigate Constitution**
  - [x] **168 - U.S. Declaration of Independence**
  - [x] **169 - The Wealth of Nations**
  - [x] **170 - Buttonwood Agreement**
  - [x] **171 - Stamp Act**
  - [x] **172 - Pride and Prejudice**
  - [x] **173 - Moonlight Sonata**
  - [x] **174 - Marcha Real**

PR #152, opened 2021-12-01
#135 - age 6 - kilo, puffing, us decl, stamp act

ultimatefiend:
-kilo,puffing,us decl, stamp act
-added function to count symbols in a pile (will be useful for heritage and supremacy later)


micahstairs (approved):
Nice work! Looks great

PR #155, opened 2021-12-03
Kahlia age4 - tortuga, east india

ultimatefiend:


micahstairs (changes requested):
I see a few entries for Delft! Can you remove that from this pull request? The full implementation for that other card is in a separate PR.

micahstairs (approved):
Looks great, thanks!

PR #157, opened 2021-12-04
Kahlia  age4 delft

ultimatefiend:
This is stupidly complex.

You have two repeating interactions (potentially).  I created a mini state machine to manage the different paths you can take. Hopefully it makes sense....

### DIFF
--- a/innovation.game.php
+++ b/innovation.game.php
@@ -8344,6 +8344,11 @@ class Innovation extends Table
                 self::setGameStateValue('auxiliary_value', $number);
                 $step_max = 1; // --> 1 interaction
                 break;
+
+            // id 146, Artifacts age 4: Delft Pocket Telescope
+            case "146N1":
+                $step_max = 1; // --> 1 interactions
+                break;
             
             // id 149, Artifacts age 4: Molasses Reef Caravel
             case "149N1":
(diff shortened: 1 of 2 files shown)


innovation.game.php
```diff
@@ -13033,6 +13115,60 @@ function stInterInteractionStep() {
                         }
                     }
                     break;
+
+                // id 146, Artifacts age 4: Delft Pocket Telescope
+                case "146N1A":
```
micahstairs, line 13120: Okay I have some suggestions to simplify this implementation.

Right now modes 0, 1, and 2 all basically do the same thing. The only difference is which cards you are allowed to select. I'd recommend merging these cases and just setting `card_id_1` to -1 if you aren't allowed to return the first drawn card, for example.

And then to simplify the remaining code, I'd recommend splitting apart the implementation into 3 interactions instead of 2.

Interaction 1: Return a card from your score pile. If you do, draw the 2 cards. If you cannot reveal any icons in common, move directly to the 3rd interaction. Otherwise move to the 2nd interaction.
Interaction 2: Return the drawn cards (after revealing, move back to the 1st interaction)
Interaction 3: Reveal one of the drawn cards

I think this approach will drastically simplify things!

micahstairs (changes requested):

innovation.game.php
ultimatefiend: I don't think you can skip an interaction.  I've tried incrementing the step and it doesn't do anything.  It just plows ahead with the next interaction.  You will see the same thing with Maldives.  It requires some hoop jumping to get it to work.

I didn't know about the card_id_1 being -1.  Yes, that will simplify that part for sure.

micahstairs: Hmm okay so I looked at the code and you're right that it won't current let you skip interactions.

However, look for the following lines:
```
// New interaction step
self::incGameStateValue('step', 1);
```

Instead of always incrementing by 1, I think we should increment it by `$step_delta` and make that variable initialized to 1 near the top of the function. Then you'd have the freedom to set `$step_delta = 2;` if you wanted to skip a step, or you could set `$step_delta = -1;` to go back a step.

micahstairs: You'll also need to update the `if ($step == $step_max)` check to `if ($step == $step_max && $step_delta > 0)`

ultimatefiend:
I created a new branch with the same content based on the latest trunk so this pull request isn't valid anymore.

PR #161, opened 2021-12-06
Kahlia age8 2 - time, earhart

ultimatefiend:


micahstairs (approved):
Awesome work! Your approach for Earhart is pretty nice.

PR #165, opened 2021-12-08
Kahlia age 10 maldives

ultimatefiend:


micahstairs (approved):
Nice work! I found a way to simplify the implementation further by skipping the interaction if we realize it's not needed after all

PR #169, opened 2021-12-11
Kahlia dolly age 10

ultimatefiend:


micahstairs (approved):
Nice work implementing this rather complex card! I made some simplifications by adding a executeDrawAndTuck function and also by using the new mechanism we added to skip interactions after they've already begun.

Issue #134, opened 2021-11-21
Extract bonus icon image from card


micahstairs:
As a first step towards implementing the Echoes expansion, let's start working on getting the graphics ready for it.

Similar to how we used ImageMagick to extract hexagon icons from cards, let's see if we can find a way to extract circle icons from cards.

As a proof of concept, let's try extracting the bonus icon from the following image:

https://github.com/micahstairs/bga-innovation/blob/main/misc/cards/Print_EchoesCards_front/Print_EchoesCards_front-002.png

When extracting the icon, be sure to include the dark yellow border around the bonus icon.

This task **must** be done using ImageMagick. Do not manually extract the icon using other tools, or else we will not be able to modify the ImageMagick command you write and make it extract the other bonus icons later.

**NOTE:** Before you begin working on this, please **assign this issue to yourself**. If you do not do this, then there's the possibility that 2 people will end up doing the same task, which is not a good use of time.

PR #171, opened 2021-12-13
Circle bonus icons

normaljosh:
See issue #134 . Extracted bonus icons from each position, with borders included. Just leaving the .pngs in imagemagick_commands/temp for now, since I'm not sure what we will want to do spritesheet-wise.


micahstairs (approved):
This is awesome! Thanks so much for doing this work.

PR #175, opened 2021-12-18
Kahlia age6 constitution, moonlight

ultimatefiend:
This is a template for seikilos regarding revealing and re-melding a card.

This will remove the splay if one is present and there are two cards in the affected pile.


micahstairs (approved):
Thanks!

Issue #136, opened 2021-11-22
Implement Age 7 cards for Artifacts Expansion


micahstairs:
This issue will track the implementation of all Age 7 cards for the Artifacts expansion. Please update this comment before working on any of these cards (if the work is unblocked) so that we don't have duplicate work being done. We no longer need to file a separate issue for each card, since a consolidated list of the cards will keep us more organized.

  - [x] **175 - Periodic Table**
  - [x] **176 - Corvette Challenger**
  - [x] **177 - Submarine H. L. Hunley**
  - [x] **178 - Jedlik's Electromagnetic Self-Rotor**
  - [x] **179 - International Prototype Metre Bar**
  - [x] **180 - Hansen Writing Ball**
  - [x] **181 - Colt Paterson Revolver**
  - [x] **182 - Singer Model 27**
  - [x] **183 - Roundhay Garden Scene**
  - [x] **184 - The Communist Manifesto**

PR #185, opened 2022-01-18
age 7 #136 - communist

ultimatefiend:


micahstairs (approved):
Thanks for tackling this tricky card! I made some simplifications (adding a second auxiliary value also helped me do this).

Issue #162, opened 2021-12-06
Implement Age 10 cards for Artifacts Expansion


ultimatefiend:
This issue will track the implementation of all Age 10 cards for the Artifacts expansion. Please update this comment before working on any of these cards (if the work is unblocked) so that we don't have duplicate work being done. We no longer need to file a separate issue for each card, since a consolidated list of the cards will keep us more organized.

  - [x] **205 - Rover Curiosity**
  - [x] **206 - Higgs Boson**
  - [x] **207 - Exxon Valdez**
  - [x] **208 - Maldives**
  - [x] **209 - Maastricht Treaty**
  - [x] **210 - Seikan Tunnel** 
  - [x] **211 - Dolly the Sheep**
  - [x] **212 - Where's Waldo?**
  - [x] **213 - DeLorean DMC-12**
  - [x] **214 - Twister**

PR #190, opened 2022-01-22
#162 - twister

ultimatefiend:
This has overlap with Communist Manifesto (some of the reveal->board transitions).

Also, there is a Javascript bug when more than two cards are revealed from your score pile.  If you refresh the page, the cards are in the reveal area though.


micahstairs:
Okay I rebased this with the latest version of main-dev (which included the Communist Manifesto changes), so before you continue working on this branch, you need to `git pull` first.

So in your proposed implementation, you are getting the player to select a color and then getting them to meld a card of that color. What do you think of doing this in one step instead of two? For the first card we meld, we let the player choose any color, and then for succeeding melds, we reduce the colors that they are allowed to choose.

What do you think of this idea? If you like it, I will add support to the 'color' option to work in this situation (right now the 'color' option is only used for choosing colors, not choosing cards).

ultimatefiend:
I think that's a brilliant idea.  Much simpler.

Do we need to add anything?  We could even return the cards to the score pile you can't choose from (i.e. the colors that have been chosen already) so its not as confusing.

ultimatefiend:
Alternatively, the user could choose a color, reveal just those colors and then choose from one of them.

That might limit the number of cards that need to be shown at a time (it could be a lot).

micahstairs:
Okay, for simplicity, how about we reveal all cards from the score pile, and then each time the player chooses a color, we return the other cards of that color.

Please leave a TODO to improve this once we have a mechanism to dump out a list of cards to the log. I think that will be a better long-term solution. But it will also involve us adding support for filtering the cards by their color (which will be relatively easy to do).

How does that sound to you?

ultimatefiend:
Sounds good.

ultimatefiend:
Much simpler now!

micahstairs (approved):
Thanks for simplifying this!

PR #193, opened 2022-01-23
Inno 186: Make card layout to more closely match real cards

lordmcfuzz:


lordmcfuzz:
Work in progress

.gitignore
```diff
@@ -1,2 +1,5 @@
 .DS_Store
 .vscode
+img/game_display0.jpg
```
lordmcfuzz, line 3: For some reason these are in my project. whether I should delete them locally or just add them to the git ignore to prevent accidental upload from a new person is something I'm not up to deciding.

innovation.js
```diff
@@ -1282,7 +1282,7 @@ function (dojo, declare) {
             
             dojo.place(div_title, dojo.body());
             
-            // Determine the font-size between 1 and 30px which enables to fill the container without overflow
```
lordmcfuzz, line 1285: I updated things to use rem, so sizes are just 1-30 now, no need to specify unit.

misc/css_generator.py
```diff
@@ -11,18 +11,19 @@
 """
 
 # File the CSS code will be written in
-dest_file = "to_be_appended_into_innovation.css"
+dest_file = ".\\misc\\to_be_appended_into_innovation.css"
```
lordmcfuzz, line 14: I moved the to_be_appened_into_innovation.css to build in misc, so that is not uploaded to bga

misc/to_be_appended_into_innovation.css
```diff
@@ -0,0 +1,962 @@
+/* Auto-generated by css_generator.py (Python script) */
```
lordmcfuzz, line 1: Probably could add a gitignore here, but it could be of worth for peer reviewing changes that are brough upon the generator.

misc/css_generator.py
lordmcfuzz: I guess this also assumes that you are all on windows...

lordmcfuzz: If there is a need I'm sure we can upgrade the script to detect file system and adjust accordingly

innovation.js
micahstairs: Sounds good!

.gitignore
micahstairs: I don't have this file but adding it to the .gitignore sounds fine to me.

misc/css_generator.py
micahstairs: Would `dest_file = "misc/to_be_appended_into_innovation.css"` work regardless of the system?

micahstairs: Actually, given that the css_generator.py script is already in misc, if you run it from that directory won't it put it in misc?

Does `".\\misc\\to_be_appended_into_innovation.css"` make it so that it always ends up in misc regardless of which directory you run the script in?

micahstairs (changes requested):
If you look closely at the resource icons in this screenshot, do you see how the borders aren't even widths?

Screen Shot 2022-01-25 at 7 25 54 PM

micahstairs:
> If you look closely at the resource icons in this screenshot, do you see how the borders aren't even widths?
> 
> Screen Shot 2022-01-25 at 7 25 54 PM

For example, changing the `background-position-x` for `.M.icon_4.square_card_icon` from -117px to -116px made that icon look better.

misc/css_generator.py
lordmcfuzz: I'll try the /

I have been running the py by right clicking it and running it, which happens to open the terminal in the project space, so it would generate it there. hmm if you were navigating to the file, then it would error out when you run it. I will see if I can find a universal solution

lordmcfuzz:
I see, I will see what I can tweak

misc/css_generator.py
lordmcfuzz: I remembered how to do py stuff. will be fixed in upcoming commit to this branch

lordmcfuzz:
The "button" like highlight of the icon is throwing me off. I think I got them all centered within the color border now. I had to use a screenshot to zoom in.

innovation.css
```diff
@@ -1441,116 +1464,131 @@
...
 }
 
 .M.color_1.square_card_icon  {
-    background-position-y: -31px;
+    background-position-y: -39px;
 }
 
 .M.color_2.square_card_icon  {
-    background-position-y: -61px;
+    background-position-y: -77px;
```
micahstairs, line 1567: I think this should be -78px. See the "Clothing" card.

innovation.css
```diff
@@ -1441,116 +1464,131 @@
...
+    height: 36px;
 }
 
 .M.icon_1.square_card_icon  {
     background-position-x: -1px;
 }
 
 .M.icon_2.square_card_icon  {
-    background-position-x: -31px;
+    background-position-x: -39px;
```
micahstairs, line 1539: I think we want -39.5px here.

innovation.css
```diff
@@ -1441,116 +1464,131 @@
...
 }
 
 .M.icon_4.square_card_icon  {
-    background-position-x: -91px;
+    background-position-x: -116px;
 }
 
 .M.icon_5.square_card_icon  {
-    background-position-x: -121px;
+    background-position-x: -155px;
```
micahstairs, line 1551: -154.5px

innovation.css
```diff
@@ -1441,116 +1464,131 @@
...
+    background-position-x: -192.5px;
 }
 
 .M.color_0.square_card_icon  {
     background-position-y: -1px;
 }
 
 .M.color_1.square_card_icon  {
-    background-position-y: -31px;
+    background-position-y: -39px;
```
micahstairs, line 1563: -39.5px

micahstairs (changes requested):
Looking pretty good! I have some pixel tweak suggestions.

innovation.css
```diff
@@ -798,14 +798,37 @@
...
+.card_effects.M > .font_size_4 > .effect.M {
+    line-height: 5px;
+}
+.card_effects.M > .font_size_5 > .effect.M {
+    line-height: 5px;
+}
+.card_effects.M > .font_size_6 > .effect.M {
+    line-height: 6px;
+}
+.card_effects.M > .font_size_7 > .effect.M {
```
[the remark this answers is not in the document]
lordmcfuzz: anything over size 7 we use what we had before on line ~799, class .effect.M {... line-height:8px...}

innovation.css
lordmcfuzz: agreed

innovation.css
lordmcfuzz: agreed

innovation.css
lordmcfuzz: agreed

innovation.css
lordmcfuzz: agreed

innovation.css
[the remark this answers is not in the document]
micahstairs: Okay great!

misc/to_be_appended_into_innovation.css
micahstairs: Yeah it seems fine to check this in. Especially since it's in misc/ now.

micahstairs (approved):
This looks amazing, thanks!

Screen Shot 2022-01-28 at 4 24 23 PM

ultimatefiend:
This does look great!  Nice work @lordmcfuzz !

Issue #191, opened 2022-01-22
Seized relics are not subtracted from the player score at the end of the game


ultimatefiend:
**NOTE:** Before you begin working on this, please **assign this issue to yourself**. If you do not do this, then there's the possibility that 2 people will end up doing the same task, which is not a good use of time.

If a relic is stolen from another person, the BGA score is not correct at the end of the game.  It doesn't get removed from their player score for some reason.

micahstairs:
Have you confirmed whether this is a frontend-only bug or is the database also not reflecting the proper number of achievements (BGA player score)?

ultimatefiend:
I think this is a bug in the back end implementation on my end.  I should be able to test a fix tonight after kiddo bedtime.

Issue #145, opened 2021-11-27
Implement Age 8 cards for Artifacts Expansion


micahstairs:
This issue will track the implementation of all Age 8 cards for the Artifacts expansion. Please update this comment before working on any of these cards (if the work is unblocked) so that we don't have duplicate work being done. We no longer need to file a separate issue for each card, since a consolidated list of the cards will keep us more organized.

  - [ ] **185 - Parnell Pitch Drop**
    - Status: NOT STARTED
  - [ ] **186 - Earhart's Lockheed Electra 10E**
    - Status: NOT STARTED
  - [ ] **187 - Battleship Bismarck**
    - Status: NOT STARTED
  - [ ] **188 - Battleship Yamato**
    - Status: NOT STARTED
  - [ ] **189 - Ocean Liner Titanic**
    - Status: NOT STARTED
  - [ ] **190 - Meiji-Mura Stamp Vending Machine**
    - Status: NOT STARTED
  - [ ] **191 - Plush Beweglich Rod Bear**
    - Status: NOT STARTED
  - [ ] **192 - Time**
    - Status: NOT STARTED
  - [ ] **193 - Garland's Ruby Slippers**
    - Status: NOT STARTED
  - [ ] **194 - '30 World Cup Final Ball**
    - Status: NOT STARTED

ultimatefiend:
This is what I had for World Cup Ball before I realized that the transfer wouldn't work. :)

(194, 1, 8, 4, 2, 2, 6, 0, 2, FALSE, 0, 'deck', 0, NULL, FALSE),

            // id 194, Artifacts age 8: 30 World Cup Final Ball
            case "194C1":
                $step_max = 1;
                break;

            case "194N1":
                $happened = true;
                do {
                    // Draw and reveal an 8.
                    $card = self::executeDraw($player_id, 8, 'revealed');
                
                    // "The single player with the highest top card of the drawn card's color achieves it, ignoring eligibility. 
                    self::transferCardFromTo($card, $player_id, 'achievements');
                    
                    // If that happens, repeat this effect."
                    $happened = false;
                } while($happened);
                self::transferCardFromTo($card, $player_id, 'hand'); // unspoken : card is placed in hand
                break;




        // id 194, Artifacts age 8: 30 World Cup Final Ball
        case "194C1A":
            // "return one of your achievements!"
            $options = array(
                'player_id' => $player_id,
                'n' => 1,
                'can_pass' => false,

                'owner_from' => $player_id,
[...]

PR #203, opened 2022-01-25
#191, #145 - world cup

ultimatefiend:
This solves a few problems.

First, it adds a new age 8 card (yay!)

Second, to add the card, the scoring needs to decrement if an achievement is returned or transferred.  This also affected relic transfers both from the current player (when they transfer it to hand) and relic steals from another player.

Third, there was a bug in the relic in hand function that should be using "type" instead of age.


micahstairs (approved):
Nice work!

I ended up simplifying the logic for '30 World Cup Final Ball by creating two SQL queries to help us. In addition to eliminating complicated PHP logic, this also drastically cuts down on the SQL queries we are issuing.

Issue #184, opened 2022-01-16
Implement Relic cards for Artifacts Expansion


ultimatefiend:
This issue will track the implementation of all Relic cards for the Artifacts expansion. Please update this comment before working on any of these cards (if the work is unblocked) so that we don't have duplicate work being done. We no longer need to file a separate issue for each card, since a consolidated list of the cards will keep us more organized.

  - [ ] **215 - Timbuktu**
    - Status: BLOCKED
    - Needs Cities Expansion
  - [ ] **216 - Complex Numbers**
    - Status: NOT STARTED (medium)
  - [ ] **217 - Newton-Wickins Telescope**
    - Status: NOT STARTED (medium)
  - [ ] **218 - Ching Shih**
     - Status: BLOCKED
     - Needs Figures expansion
  - [ ] **219 - Safety Pin**
     - Status: BLOCKED
     - Needs Echoes expansion

micahstairs:
Thanks for filing this! I removed 3 of the relics from this issue to avoid confusion.

ultimatefiend:
Did you create an issue for the other expansions so we don't forget?

micahstairs:
No, but I have a personal list on my phone to remind me that we need to consider relics when it's time to implement them. I'd prefer not to clutter up our list of tasks for now!

ultimatefiend:
is complex numbers unblocked?

micahstairs:
Oh sorry, it's still blocked. I re-opened that other issue. I accidentally closed it early.

PR #206, opened 2022-01-26
#184 - newton wickins relic

ultimatefiend:
this also fixes complex number's demand.  It is now a non-demand.


micahstairs (approved):
Great work! I didn't even have any tweaks I wanted to make.

PR #207, opened 2022-01-26
#136 - colt paterson

ultimatefiend:


micahstairs (approved):
Awesome! I added a TODO because I think we can improve this card later by dumping out the list to the game log instead of moving each card to the revealed area and then putting it back in hand.

Issue #168, opened 2021-12-10
Create "board->removed" and "hand->removed" transitions


ultimatefiend:
**NOTE:** Before you begin working on this, please **assign this issue to yourself**. If you do not do this, then there's the possibility that 2 people will end up doing the same task, which is not a good use of time.

Delorean needs to remove cards from the board and hand from the game.  Exxon will need to do something similar.

Right now, there is no transition for removed cards as Fission has its own function call that removes its cards from the game all in one go.

ultimatefiend:
Big Bang needs to do something similar. (draw and remove a 10)

micahstairs:
So I don't think we want to add a hand->removed transition because I think we will want a custom method to remove all hands at once (otherwise animations will take too long I believe). In fact I'll probably refactor Fission to use this method too. And then I'd make another separate method for removing all score piles, and another for all achievements, etc.

ultimatefiend:
sure, that makes sense.  Big bang does need a hand-> removed transition for a single card though.

micahstairs:
Some of these transitions were added with the initial implementation of Exxon Valdez, however, there are some changes to be made to the card, so I've left TODOs to remove these transitions afterwards.

I'm going to close this though.

ultimatefiend:
FYI, hand->removed will need to remain when figures is implemented since you remove your hand from the game when you trigger the figures special achievements.

micahstairs:
Good to know. We can re-add that if we ever manage to implement Figures :)

For now it's best to keep the number of strings that need to be translated to a minimum.

PR #214, opened 2022-01-29
#162, #168 - exxon

ultimatefiend:
This is the first card that can remove a player!!

This also makes demand effects not work on eliminated players.


micahstairs (approved):
Thanks for working on this!

I'm going to approve and merge it, however, there's still some work to be done on this card. Some of the TODOs are for the making some changes to the log messages, feel free to address those in another PR. However, for the other TODOs involving removing all of the cards at once, I'll file an issue for that since there's also some refactoring to do with the Fission card.

PR #223, opened 2022-02-04
#162 - rover

ultimatefiend:


micahstairs (approved):
Haha that was an easy card to implement!

PR #230, opened 2022-02-06
Cross of Coronado - remove TODO 

ultimatefiend:
make the reveal automatic rather than an interaction (removing a TODO)


micahstairs (approved):
Thanks for making this improvement!

PR #234, opened 2022-02-08
INNO-205: icon hash tracked by global variables

lordmcfuzz:


lordmcfuzz:
fyi, this code has errors, at this point I'm wondering if I understand what I am trying to solve here.

micahstairs:
Okay so first of all, global variables must be between 10-99. You used indices reserved for game options.

Secondly, I'd recommend searching the code for `card_id_1` and getting a sense of how that is used to restrict the card selection to cards with that ID. We are going to do something similar here by restricting the card selection to cards with a particular hash

micahstairs:
Also in your SQL query, make sure to wrap the ORs in parentheses:

```
AND (icon_hash = {icon_hash_1} OR icon_hash = {icon_hash_2} OR icon_hash = {icon_hash_3} OR icon_hash = {icon_hash_4} OR icon_hash = {icon_hash_5})
```

micahstairs:
Closing this because I submitted the full version of these changes already.

PR #237, opened 2022-02-09
INNO-194: resize action card to be inline with icons

lordmcfuzz:


lordmcfuzz:
I did not change the action_text to be smaller. I think it looks ok, If I shrink the box the text is in we will drop ~2 font sizes and it looks weird.

micahstairs:
This looks better!

Screen Shot 2022-02-09 at 4 48 34 PM

But I think the spacing between lines of text is a little too big. Can you also shrink the line-height from 15px to 12px and then fix the vertical alignment?

lordmcfuzz:
Suggestion to review "Second Action" too

micahstairs:
Screen Shot 2022-02-09 at 9 44 19 PM

Screen Shot 2022-02-09 at 9 41 59 PM

Screen Shot 2022-02-09 at 9 46 24 PM

My only critique is that we are using a different front size for "Second Action". However, I think we need dynamic font sizes in order to gracefully handle translations, right?

micahstairs:
Ooh what if we set the max font size? Let's pick the max to be font_size_12, since we are using that for "Second Action". Thoughts?

micahstairs:
In givePlayerActionCard we are currently setting the max font size to be 15:

```
var div_action_text = this.createAdjustedContent(action_text, 'action_text', '', 15, 2);
```

lordmcfuzz:
Yeah Dynamic Text size is needed here(like most places) and that prevented me from using display:flex features as the dynamic text size picker goes immediately to 1.

I'm good with setting the max to 12. It will make the english match but any translation is not guaranteed to match.

micahstairs (approved):
This looks perfect!

PR #242, opened 2022-02-11
INNO-241: Setting hex icon classes for relics

lordmcfuzz:
I'm fairly sure this works but is currently untested as relics may be bugged.


micahstairs (approved):
Thanks!

Issue #202, opened 2022-01-25
Add Artifact-specific statistics


ultimatefiend:
**NOTE:** Before you begin working on this, please **assign this issue to yourself**. If you do not do this, then there's the possibility that 2 people will end up doing the same task, which is not a good use of time.

At the end of every game, there is a list of statistics listed below the results section:

![image](https://user-images.githubusercontent.com/11416698/150895878-c3b304ad-445e-4ab2-afb5-756c1b2b926f.png)

Should we add a section for artifacts?  If so, what should they be?

For example, "relics seized" or "relics lost" might be good ones.

micahstairs:
It would also be nice to know how the free action is being used for Artifacts on display.

ultimatefiend:
More info:

https://en.doc.boardgamearena.com/Game_statistics:_stats.inc.php

I didn't see a way to hide stats.

micahstairs:
There's more info on this page: https://en.doc.boardgamearena.com/Main_game_logic:_yourgamename.game.php

```
initStat( $table_or_player, $name, $value, $player_id = null )

Create a statistic entry with a default value.

This method must be called for each statistic of your game, in your setupNewGame method. If you neglect to call this for a statistic, and also do not update the value during the course of a certain game using setStat or incStat, the value of the stat will be undefined rather than 0. This will result in it being ignored at the end of the game, as if it didn't apply to that particular game, and excluded from cumulative statistics. As a consequence - if do not want statistic to be applied, do not init it, or call set or inc on it.
```

micahstairs:
There's also some work here to make sure existing statistics don't break. For example, we should make sure I demand effects aren't broken by the new I compel mechanism. There are also a few statistic-related TODOs in the code right now which need to be addressed as part of this issue.

micahstairs:
I'll finish the outstanding 3 TODOs!

PR #269, opened 2022-02-21
#202 - stats

ultimatefiend:
This is still missing logic for the demand/compel statistics.

I'm not sure where they go.  Presumably they would end up in the nested execution logic?


micahstairs (approved):
Sorry it took me a while to look at this!

Made a few tweaks to the wording, as well as moving some of the stat incrementing so that it gets called before the player potentially ends the game by getting enough achievements.

micahstairs:
Also left a few TODOs for the outstanding work.

Issue #279, opened 2022-02-22
Automate card transfers in Ship of the Line Sussex


micahstairs:
**NOTE:** Before you begin working on this, please **assign this issue to yourself**. If you do not do this, then there's the possibility that 2 people will end up doing the same task, which is not a good use of time.

"And score all cards of that color from your board" shouldn't be an interaction. It should be an automated step that happens during 158N1B after the color is chosen.

Issue #278, opened 2022-02-22
Automate card transfers in Mjolnir Amulet


micahstairs:
**NOTE:** Before you begin working on this, please **assign this issue to yourself**. If you do not do this, then there's the possibility that 2 people will end up doing the same task, which is not a good use of time.

Right now the player has to score cards one at a time from their board. We should automate this. We should also modify the first interaction so that we are actually just choosing a top card (instead of transferring it to the score card right away). This will lead to wording in the game log that is more consistent with the dogma effect text.

PR #287, opened 2022-02-23
#278, #279 - automate mjolnir and sussex

ultimatefiend:


micahstairs (approved):
Thanks! I made a change to score from top to bottom instead of bottom to top.

PR #314, opened 2022-03-08
Inno 252: SCSS upgrade

lordmcfuzz:
I updated the readme to reflect the additional set up so that the scss compilation is automatic as well as the python script running.


innovation.scss
```diff
@@ -0,0 +1,4482 @@
...
+.font_size_29 {
+	font-size: 1.8125rem;
+}
+
+.font_size_30 {
+	font-size: 1.875rem;
+}
+
+.M { 
+	&.hexagon_card_icon {
```
micahstairs, line 3184: Why is hexagon_icon_0 defined 4 times? I think we have all of the hex icons defined twice (which is thousands of lines of code)

micahstairs (changes requested):

innovation.scss
lordmcfuzz: It was an issue with me not manually deleting a section of code and then I changed the string on the css generator and so it added it's output to the end of the old stuff that did not get deleted.

So the correct number is two and not 4. Though getting it down to one would be cool and at the top of my head if I generated everything in a json format and then wrote the json to scss it could refactor that down. Not sure how much of a benefit that would be right now

micahstairs: Adding another level on top of this seems like overkill!

lordmcfuzz: It would be run the loops and build a large json object -> feed that json object into a scss printer. so instead of two sizes of .[size]&.[class1]&.[class2] { code } it would be .[class1]&.[class2] { &[size1] {code} &.[size2] {code} }

misc/css_generator.py
```diff
@@ -11,9 +11,10 @@
 """
 
 import os
+import mmap
 
-# File the CSS code will be written in
-dest_file = "to_be_appended_into_innovation.css"
+# File the SCSS code will be written in
+dest_file = "..\innovationlordmcfuzz.scss"
```
micahstairs, line 17: I see lordmcfuzz in here. Does this mean that the rename script wasn't able to change this?

innovation.scss
```diff
@@ -0,0 +1,3059 @@
...
+		}
+	}
+}
+/* scss_generator code starts here */
+/* anything below this will get blown away on rebuild */
+
+.font_size_1 {
+	font-size: 0.0625rem;
+}
+
```
micahstairs, line 1643: Can't we do something like this here? https://stackoverflow.com/questions/50699726/calculation-based-on-class-name-in-sass

I expect we can also do something similar for the hexagons. With SCSS, might we be able to eliminate the Python script altogether?

innovation.scss
```diff
@@ -0,0 +1,3059 @@
...
+		>.font_size_4 {
+			>.effect.M {
+				line-height: 5px;
+			}
+		}
+		>.font_size_5 {
+			>.effect.M {
+				line-height: 5px;
+			}
+		}
```
micahstairs, line 775: Can't we condense this?

innovation.scss
```diff
@@ -0,0 +1,3059 @@
...
+		height: 69px;
+		background-size: 100px auto;
+		background-position: -2px -145px;
+	}
+	&.side_1.M {
+		background-position: -2px -2px;
+	}
+	&.side_2.M {
+		background-position: -2px -408px;
+	}
```
micahstairs, line 232: It might be more natural to put both of these inside .M above.

micahstairs (changes requested):

innovation.scss
lordmcfuzz: yeah, I did not see it before but yeah.

innovation.scss
lordmcfuzz: OOO yes we can, I'll look to do that

misc/css_generator.py
lordmcfuzz: I missed that in the readme and realize it is probably not great. I'm looking into eliminating the py script so it won't matter.

But you are correct, likely because it is in a file and not renaming the file itself.

innovation.game.php
```diff
@@ -106,7 +106,6 @@ function __construct()
             'icon_hash_3' => 81,
             'icon_hash_4' => 82,
             'icon_hash_5' => 83,
-            'enable_autoselection' => 84,
```
lordmcfuzz, line 109: hmmm I thought I pulled up to main-dev

misc/css_generator.py
lordmcfuzz: No more py

innovation.game.php
lordmcfuzz: I did and I'm not sure how I did this.

innovation.css.map
```diff
@@ -0,0 +1 @@
...AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AASF;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC;;AAJD;EACC;;AAED;EACC","file":"innovationlordmcfuzz.css"}
```
micahstairs, line 1: Is this file supposed to be added to gitignore? Also, I see "innovationlordmcfuzz" in there.

innovation.scss
```diff
@@ -0,0 +1,1721 @@
...
+		height: 15px;
+	}
+	&.L {
+		left: 143px;
+		top: 18px;
+		width: 230px;
+		height: 37px;
+	}
+}
+.i_demand {
```
micahstairs, line 723: These three can be grouped together.

innovation.scss
```diff
@@ -0,0 +1,1721 @@
...
+			&.L {
+				background-position: -#{$L_x_position + ($i + 5) % 15 * $L_offset}px -#{$L_y_position + ((math.floor(math.div(($i - 100),10)))) * $L_offset}px;
+			}
+		}
+	}
+}
+
+
+// /* scss_generator code starts here */
+// /* anything below this will get blown away on rebuild */
```
micahstairs, line 1720: Is this obsolete now?

micahstairs (changes requested):

innovation.css.map
lordmcfuzz: I have to do some research into it. I think it is required but that is without actually knowing why. If it is I'll have to do some sort of solution in the sharecode, which would improve it for them.

innovation.scss
lordmcfuzz: Yes. Ooops

innovation.scss
lordmcfuzz: yup!

innovation.css.map
lordmcfuzz: It looks like it is useful for dev purposes, so I think it should just be ignored and it will be generated locally when you actually need it:
https://css-tricks.com/should-i-use-source-maps-in-production/

I will add to gitignore

README.md
```diff
@@ -60,7 +63,29 @@ Then navigate to https://github.com/elaskavaia/bga-sharedcode/pulls, click "New
...
+        {
+            "match": ".*\\.scss$",
+            "command": "sass ${file} ${fileDirname}/${fileBasenameNoExtension}.css",
+            "runIn": "backend",
+            "runningStatusMessage": "Compiling ${fileBasename}",
+            "finishStatusMessage": "${fileBasename} compiled"
+        }
+    ]
+```
+- Every time you edit and save in the py generator or the scss file, it will compile into css read to be used.
```
micahstairs, line 90: This still references the python file.

innovation.scss
```diff
@@ -0,0 +1,1701 @@
...
+		&.bottom_right_icon {
+			bottom: 23px;
+			right: 26px;
+		}
+	}
+}
+
+/* Hexagon icons take up ~27 % of a real cards width. 
+*  Our cards will shoot for 25 % ->  M.45px
+*  If you ever change the background-size, you must edit the offset in css_generator.py
```
micahstairs, line 1492: This still references the python file.

innovation.scss
```diff
@@ -0,0 +1,1701 @@
...
+$rem: 0.0625rem;
+@for $i from 1 through 30 {
+	.font_size_#{$i} {
+		font-size: $rem * $i;
+	}
+}
+
+/* hexagon card icons */
+
+$M_offset:		49;
```
micahstairs, line 1635: This entire file seems to be using tabs. Please use spaces!

README.md
```diff
@@ -7,9 +7,12 @@ In order to test changes to the game, you need to be able to copy the code out o
 #### Prerequisites
 - Create a BGA developer account on https://studio.boardgamearena.com/.
 - Create your own BGA Studio project for Innovation at https://studio.boardgamearena.com/studio. Make sure to suffix the new project name with your username (e.g. `InnovationJohnSmith`).
+- Optionally Install github desktop (https://desktop.github.com/) -> this is a simplier way of using git if you want to skip the commandline
```
micahstairs, line 10: simpler

README.md
```diff
@@ -7,9 +7,12 @@ In order to test changes to the game, you need to be able to copy the code out o
...
 - Clone this repository onto your computer (`git clone https://github.com/micahstairs/bga-innovation.git innovation`). You'll make your life easier if the directory is named `innovation` instead of `bga-innovation`.
 - Clone https://github.com/elaskavaia/bga-sharedcode onto your computer (`git clone https://github.com/elaskavaia/bga-sharedcode.git`).
 - Install PHP (https://windows.php.net/download#php-7.3).
+- Install SASS (https://sass-lang.com/install)
```
micahstairs, line 14: Maybe add a note that this is only needed if they are going to make changes to SCSS.

README.md
```diff
@@ -60,7 +63,29 @@ Then navigate to https://github.com/elaskavaia/bga-sharedcode/pulls, click "New
...
+    ],
+    "syncOption": {
+        "skipCreate": false,
+        "delete": true 
+    }
 }
 ```
+- Sometimes you will have to run a manual push of your files and you can do that by using the command shortcut `Ctrl+Shift+P` and runing `SFTP: Sync Local -> Remote`
+- Install Run on Save extension (https://marketplace.visualstudio.com/items?itemName=pucelle.run-on-save)
+- Configure the Run on Save extention by clickin the Extensions tab in VSCode -> Gear icon on Run on Save -> Extension Settings -> Edit settings.json -> copy this into that file
```
micahstairs, line 78: clicking

README.md
```diff
@@ -60,7 +63,29 @@ Then navigate to https://github.com/elaskavaia/bga-sharedcode/pulls, click "New
         ".vscode",
         ".git",
         ".DS_Store",
-        "misc/"
-    ]
+        "misc/",
+        "node_modules",
+        "package-lock.json"
+    ],
+    "syncOption": {
```
micahstairs, line 70: What does this do?

README.md
micahstairs, line 78: Can you add a note that these two steps are only needed if they want to edit SCSS? Most developers won't need to touch this.

micahstairs (changes requested):
Just a few small comments, this is looking good!

I expect there are further optimizations we can make to the SCSS but they don't need to be made in this PR.

innovation.scss
lordmcfuzz: That is likely a vscode setting, I will update and hopefully use a short cut to replace

README.md
lordmcfuzz: ah yes I think I lost that note to re-review this file. sorry, embaressing.

README.md
lordmcfuzz: https://github.com/liximomo/vscode-sftp/wiki/config#syncoption

deletes files that no longer are at the origin point and creates not files at destination

micahstairs (approved):
Thanks, this looks good! I haven't tried making changes to the SCSS changes myself yet, but I plan on making a few small changes in the near future.

Issue #329, opened 2022-03-23
Holy Lance looks for Holy Grail on any board, not the player's board


ultimatefiend:
**NOTE:** Before you begin working on this, please **assign this issue to yourself**. If you do not do this, then there's the possibility that 2 people will end up doing the same task, which is not a good use of time.

$card = self::getIfTopCardOnBoard(131);

This should be more selective, something like this:

 $top_card = self::getTopCardOnBoard($player_id, $color);

Where color is yellow and then compare the id of the card to 131.

PR #333, opened 2022-03-26
#329 - fix holy lance

ultimatefiend:


micahstairs (approved):
Thanks for finding and fixing this bug!

PR #337, opened 2022-03-28
fix gujin when sharing with a person with no yellow

ultimatefiend:


micahstairs (approved):
Thanks for finding and fixing!

PR #346, opened 2022-04-03
INNO-254: change age squares from image to css

lordmcfuzz:

PR #350, opened 2022-04-04
Inno 321 fix achievement width

lordmcfuzz:
dependent on #346

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,7 @@ Then navigate to https://github.com/micahstairs/bga-innovation/pulls, click "New
     "runOnSave.commands": [
             {
                 "match": ".*\\.scss$",
+                "notMatch": "variables.scss$",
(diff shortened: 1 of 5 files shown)


micahstairs (changes requested):
When I resize my window on desktop, I can see the achievement zone changing size, but then I also see the reference card being cut off. Do you know why this is happening?

lordmcfuzz:
So allot of our calculations are done on page load,  and from what I can tell this is online with bga games in general, so the whole responsive design that would be held to a standard of window resizing and the page perfectly reacting on the fly may not be easily achieved. I think we should shot for it, but incrementally. In classic bga fashion, "hit f5" when you have your page sized the way you want.

micahstairs:
Okay that is a fair point!

Now that the other PRs are merged in, can you update this one so that I have a clean diff to review?

micahstairs (approved):
Thanks!

Issue #353, opened 2022-04-09
Transcribe all Echoes cards


ultimatefiend:
This converts the raw images into strings in the materials.inc.php file.

There are 110 cards in total (including the special achievements.  The IDs will span from 330-439.

PR #365, opened 2022-04-10
#353 - all echoes cards

ultimatefiend:


ultimatefiend:
This would have no effect on the current release, but would understand if you want to wait. :)

micahstairs (approved):
Thanks! I corrected some things, but overall you didn't make many mistakes!

PR #395, opened 2022-04-17
INNO-386 + 380: icon and title updates

lordmcfuzz:
Also fixed some color issues with the tooltip icon counts


micahstairs (approved):
This is definitely an improvement, thanks!

PR #405, opened 2022-04-24
Cities - onmeld

ultimatefiend:
Fixed some TODOs for Cities
Added game options for Echoes and Cities


micahstairs (approved):
Thanks for getting this work started! I fixed a few bugs (visually), but obviously we can do playtesting later once we get more functionality in place.

Issue #367, opened 2022-04-10
Implement Age 1 cards for Echoes Expansion


ultimatefiend:
This issue will track the implementation of all cards from this age for the Echoes expansion. Please update this comment before working on any of these cards (if the work is unblocked) so that we don't have duplicate work being done.

  - [x] **330- Dice**
  - [x] **331- Perfume**
  - [x] **332- Ruler**
  - [x] **333- Bangle**
  - [x] **334- Candles**
  - [x] **335- Plumbing** (BLOCKED)
  - #497
  - [x] **336- Comb**
  - [x] **337- Ice Skates**
  - [x] **338- Umbrella** 
  - [x] **339- Chopsticks** (BLOCKED)
  - #420
  - [x] **340- Noodles**
  - [x] **341- Soap**
  - [x] **342- Bell**
  - [x] **343- Flute**
  - [x] **344- Puppet**

ultimatefiend:
Most of these cards will be blocked because the basic functions of echoes aren't present at the moment, but this is here for when they are. :)

PR #421, opened 2022-04-29
#367 - bangle, bell - some bug fixes

ultimatefiend:


micahstairs (approved):
Nice! I test it and the cards work (minus the Echo effects). I left TODOs to test them later.

PR #423, opened 2022-04-30
Kahlia echoes age1 2 - plumbing + umbrella

ultimatefiend:


micahstairs (approved):
I skimmed it quickly and this seems fine. Later on, when we I'm focusing more of my time on Echoes, I'll probably give all of the cards a second look.

Issue #417, opened 2022-04-27
Implement Age 2 cards for Echoes Expansion


ultimatefiend:
This issue will track the implementation of all cards from this age for the Echoes expansion. Please update this comment before working on any of these cards (if the work is unblocked) so that we don't have duplicate work being done.

Almost all of these cards are blocked visually by the lack of bonus icons on the cards. They look like crowns right now.
https://github.com/micahstairs/bga-innovation/issues/354

  - [x] **345- Lever** (BLOCKED)
  - #436
  - #472 
  - [x] **331- Linguistics** (BLOCKED)
  - #472
  - [x] **332- Crossbow** (BLOCKED)
  - #445
  - [x] **333- Horseshoes**
  - [x] **334- Glassblowing** 
  - [x] **335- Scissors** (BLOCKED)
  - #497
  - [x] **336- Toothbrush**
  - #420
  - [x] **337- Watermill**
  - [x] **338- Pagoda**
  - [x] **339- Chaturanga**

PR #424, opened 2022-04-30
#417 - linguistics

ultimatefiend:


micahstairs (approved):
I simplified getVisibleBonusesOnBoard a bit using array_merge.

PR #428, opened 2022-05-01
Add ability to transfer cards from bottom of deck to available achievements (fixes #420)

### LINKED ISSUE #420: Transferring cards to available achievements isn't supported
I tried transferring a card from a deck to the available achievements and I found nothing but issues.

In particular, Chopsticks and Toothbrush need to transfer the *bottom* card of a deck to the available achievements so that should work as well.

micahstairs:
Tested by calling this:

```
self::executeDraw(0, 1, 'achievements', /*$bottom_to=*/ false, /*type=*/ 0, /*bottom_from=*/ true);
```

Screen Shot 2022-05-01 at 12 03 17 PM

### DIFF
--- a/innovation.game.php
+++ b/innovation.game.php
@@ -1319,7 +1319,7 @@ class Innovation extends Table
     /**
      * Executes the transfer of the card, returning the new card info.
      **/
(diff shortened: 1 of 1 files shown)


ultimatefiend (approved):
I think the changes look good.  I will try it out with Chopsticks once they are pushed.

PR #430, opened 2022-05-01
Some quality of life improvements on the config file

splittingred:
First-time PR here! Few tiny little things that were bugging me as I was digging in:

* Using PHP shorthand for arrays (preferred since PHP 5.4+): `[` `]`
* Adding some documentation links for tags to cross-reference, and commenting what each scalar value is
* Adding a `presentation` block (from bgg page) to make the game page a little nicer.

Let me know if this isn't desired or I'm missing something, or this shouldn't be filed against this (`main-dev`) branch! (Or against another repo?)

### DIFF
--- a/gameinfos.inc.php
+++ b/gameinfos.inc.php
@@ -1,6 +1,6 @@
 <?php
 
-$gameinfos = array( 
+$gameinfos = [
 
 // Game designer (or game designers, separated by commas)
 'designer' => 'Carl Chudyk',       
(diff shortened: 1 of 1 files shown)


gameinfos.inc.php
```diff
@@ -76,7 +76,19 @@
...
+    3, // for regular players
+    12, // long game >30m
+    20, // awarded game
+    102, // historical
+    106, // building
+    200, // cards
+    207, // combos
+    208, // area majority
+    209, // race
+    210 // collection
```
micahstairs, line 90: To make future additions less error-prone, what do you think of adding a comma after 210?

We aren't doing this consistently in our PHP files yet, but I'm trying to move us in that direction incrementally.

micahstairs (changes requested):
Thanks! Just requesting one small change.

gameinfos.inc.php
splittingred: Happy to! Added.

micahstairs (approved):
Thanks!

PR #431, opened 2022-05-02
Add PHPUnit and example tests, github action to run them automatically

splittingred:
* Adds composer and phpunit. This allows for writing tests against code locally
* Adds a github action to run tests on PRs + merge, which will allow for continuous integration
* Adds https://github.com/danielholmes/bga-workbench which has some nice utilities for testing BGA games in phpunit

Note I added `vendor/` to the gitignore list - this is intentional as we don't want to sync in all test dependencies into git.

Note also that without doing [something like this](https://github.com/danielholmes/bga-workbench#compiling-composer-projects) we shouldn't use composer dependencies in the _actual_ game, as that won't be on the actual servers.


tests/Unit/Utils/StringsTest.php
```diff
@@ -0,0 +1,31 @@
...
+    {
+        return [
+            ['abacus', 'zebra', true], # alpha precedence
+            ['zebra', 'abacus', false], # alpha precedence, reverse case
+            ['1vision', 'abacus', true], # numbers prior to letters
+            ['aaaa', '1abc', false], # numbers prior to letters, reverse case
+            ['aaa', 'abbb', true], # alphabet precedence for second letter
+            ['aaa', 'a11', false], # second numbers over letters
+            ['A Big Card', 'ABC Card', false], # spaces _after_ letters
+            ['Code of Laws', 'Pottery', true], # Real example
```
micahstairs, line 28: Can you add a real example using accents? e.g. `São Paulo`

.gitignore
```diff
@@ -1,8 +1,10 @@
 .DS_Store
+.idea
```
micahstairs, line 2: Thanks for updating .gitignore but can you also update the sample SFTP config in the readme to avoid uploading these files to BGA too?

micahstairs (changes requested):
Just requesting a couple small changes! Otherwise this all looks good I think. I haven't tested out Composer myself yet, but I'm glad it can run on push and pull requests.

tests/Unit/Utils/StringsTest.php
micahstairs: According to Chris, we should treat accented characters as if they are not accented. So ideally they are treated as equals.

micahstairs (changes requested):

tests/Unit/Utils/StringsTest.php
splittingred: So then who goes first if the card is identical but one has accents?

micahstairs: There aren't any other cards starting with Sao, and it's possible that's the only accent in the titles.

Can we make sure Sailing comes before São Paulo, and Sao Paulo comes before Satellites? If that works, I think we should be good.

splittingred:
Clean test run: https://github.com/splittingred/bga-innovation/actions/runs/2290967588

Had to do some tweaking to the test suite to 1) have it know it was running in GitHub Actions and hardcode a name, and 2) allow local dev to not have to rename classes to work.

README.md
```diff
@@ -65,7 +68,8 @@ Then navigate to https://github.com/micahstairs/bga-innovation/pulls, click "New
         ".DS_Store",
         "misc/",
         "node_modules",
-        "package-lock.json"
+        "package-lock.json",
+        "vendor/"
```
micahstairs, line 72: Please add `composer.json`, `composer.lock`. `phpunit.xml` and `tests/`.

modules/Utils/Strings.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+namespace Utils;
```
micahstairs, line 3: Did you see the modules README?

```
This directory is meant to hold PHP / js files for inclusion in your
<gamename>.game.php / gamename.js if needed.

It is recommended to name your files starting with a trigram of your game name.
For example if you have a class to manage pathfinding in the game
Tobago, it will be called TBGPathFinding.php (and the class inside be
called TBGPathFinding) to avoid potential namespace collisions.
```

Let's chat about this.

micahstairs (changes requested):

modules/Utils/Strings.php
splittingred: I've updated it to `Innovation` with namespaces, but we're pending the answer to https://discord.com/channels/753304735615811584/753750063725150289/973041581349285980 to see if we need to add some kind of improved syncing+renaming tooling here to handle this more fluidly for forks.

micahstairs (approved):
Thanks! I'm super excited to have a testing framework set up for this project.

PR #443, opened 2022-05-11
Add GameState class for easy and testable access of game state

splittingred:
* Adds a utility class for accessing game state, rather than using static methods (which are generally difficult to test). This allows this class to eventually be mockable and injectable (passed into a constructor).
* Adds an "arrays" utility class for various encoding/manipulation of array data
* Upgrades PHPUnit to 8, which adds better mocking functionality.

This progresses toward paring down the size of the main game class, which is huge and hard to manage. It allows this codebase to take a more SRP ([Single Responsibility Principle](https://en.wikipedia.org/wiki/Single-responsibility_principle)) approach toward modeling, which will enable easier and more assured refactoring with less risk of bugs.

Note I didn't replace _all_ instances of the static methods around sate in the Innovation codebase; just a couple low-risk ones for now. Replacing all would make this PR _huge_ and I felt it best to keep things small at first.

### DIFF
--- a/innovation.game.php
+++ b/innovation.game.php
@@ -18,13 +18,18 @@
 
 
 require_once(APP_GAMEMODULE_PATH.'module/table/table.game.php');
+require_once('modules/Innovation/Utils/Arrays.php');
 require_once('modules/Innovation/Utils/Strings.php');
+require_once('modules/Innovation/GameState.php');
 
 /* Exception to be called when the game must end */
 class EndOfGame extends Exception {}
 
 class Innovation extends Table
 {
+    /** @var \Innovation\GameState An inverted control structure for accessing game state in a testable manner */
+    private \Innovation\GameState $innovationGameState;
+
     function __construct()
     {
         // Your global variables labels:
@@ -34,6 +39,7 @@ class Innovation extends Table
         //  the corresponding ID in gameoptions.inc.php.
         // Note: afterwards, you can get/set the global variables with getGameStateValue/setGameStateInitialValue/setGameStateValue
         parent::__construct();
+        $this->innovationGameState = new \Innovation\GameState($this);
         self::initGameStateLabels(array(
             'number_of_achievements_needed_to_win' => 10,
             'turn0' => 11,
@@ -181,7 +187,7 @@ class Innovation extends Table
             }
         }
 
-        if (self::getGameStateValue('release_version') == 0) {
+        if ($this->innovationGameState->get('release_version') == 0) {
             self::initGameStateLabels(array(
                 'card_id_1' => 69,
                 'card_id_2' => 70,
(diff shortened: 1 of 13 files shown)


innovation.game.php
```diff
@@ -18,13 +18,18 @@
...
 require_once('modules/Innovation/Utils/Strings.php');
+require_once('modules/Innovation/GameState.php');
 
 /* Exception to be called when the game must end */
 class EndOfGame extends Exception {}
 
 class Innovation extends Table
 {
+    /** @var \Innovation\GameState An inverted control structure for accessing game state in a testable manner */
+    private \Innovation\GameState $innState;
```
micahstairs, line 31: I'm not sold on the naming of this variable. Is there a particular reason you didn't use `gameState`?

modules/Innovation/GameState.php
```diff
@@ -0,0 +1,73 @@
...
+        return $this->game->incGameStateValue($key, $value);
+    }
+
+    /**
+     * @param string $key
+     * @param $value
+     * @return void
+     */
+    public function setInitial(string $key, $value)
+    {
```
micahstairs, line 52: Why do you sometimes put the brace on the next line and other times you put it on the same line? For consistency, let's put it on the same line.

micahstairs (changes requested):
I like the direction that you are heading with this refactoring!

innovation.game.php
splittingred: Yes, since this extends an internal BGA class - `Table` - I wanted guard against them potentially changing things internally, so prefixing with an Innovation-specific name. There's a _very_ good chance they already have a `$gameState` instance variable on the class this would break.

modules/Innovation/GameState.php
splittingred: A miss on my part, need to enable auto-formatting in my IDE for this project. Fixed!

innovation.game.php
splittingred: Would something like `innGameState` be preferred? I'm hesitant to add any vanilla instance variables here without a prefix.

micahstairs: Makes sense. How about "innovationGameState"? When I read "inn", it doesn't translate to "innovation" in my head

tests/Unit/Innovation/Utils/ArraysTest.php
```diff
@@ -0,0 +1,64 @@
...
+    {
+        $this->expectException(\BgaVisibleSystemException::class);
+        $this->expectExceptionMessage('setGameStateBase16Array() cannot encode integers smaller than 0 or larger than 15');
+        Arrays::getValueFromBase16Array([1,16,2]);
+    }
+
+    /**
+     * @dataProvider providerTestGetBase16ArrayFromValue
+     */
+    public function testGetBase16ArrayFromValue($expected, $inputValue)
```
micahstairs, line 54: I think it would be valuable adding another test which confirms that we can setGameStateBase16Array and then getBase16ArrayFromValue and end up with the same value that we started with.

modules/Innovation/GameState.php
micahstairs: Do you feel strongly about putting braces on the next line instead of the same line? It goes against the style we've been using thus far.

splittingred: Yep, I do - it follows PSR standards and is what the PHP community at-large follows. Not doing so is actually out of step with broader PHP standards: 
- https://www.php-fig.org/psr/psr-2/#1-overview
- https://www.php-fig.org/psr/psr-12/#44-methods-and-functions

micahstairs (approved):
Thanks!

PR #444, opened 2022-05-11
Improve alphanumeric card name checks to handle case insensitivity and punctuation

splittingred:
* Improve alphanumeric card name checks to handle case insensitivity and punctuation
* Adds .gitignore (and sftp) entry for ignoring phpunit's result cache file


micahstairs (approved):
Thanks for fixing this up!

Issue #446, opened 2022-05-12
Implement Age 3 cards for Echoes Expansion


ultimatefiend:
This issue will track the implementation of all cards from this age for the Echoes expansion. Please update this comment before working on any of these cards (if the work is unblocked) so that we don't have duplicate work being done.

Almost all of these cards are blocked visually by the lack of bonus icons on the cards. They look like crowns right now.
https://github.com/micahstairs/bga-innovation/issues/354

  - [x] **355- Almanac** 
  - [x] **356- Magnifying Glass**
  - [x] **357- Liquid Fire**
  - [x] **358- Katana**
  - [x] **359- Charitable Trust**
  - [x] **360- Homing Pigeons** (BLOCKED)
  - #447
  - [x] **361- Deodorant** 
  - [x] **362- Sandpaper**
  - [x] **363- Novel**
  - [x] **364- Sunglasses**
  - #449

PR #450, opened 2022-05-14
#446 - a biggie - almanac, magnifying glass, liquid fire, katana, cha…

ultimatefiend:
This is a big PR.

It has 8 new cards.  The only age 3 card left after this is sunglasses and that's due to the complexity of its dogma.

A few notes on the content here:
1. Almanac's forecast return mechanism doesn't totally work, but the logic is sound I think.  There is an issue for this.
2. Homing Pigeon's score returning logic doesn't work.  There is an issue for this as well.
3. Charitable Trust is not testable until echoes are working.  The logic *should* work though.  
4. Novel's log message when Supremacy is already claimed doesn't show the card name and I have no idea why.


micahstairs:
I just fixed the 4th item you mentioned:
Screen Shot 2022-05-14 at 10 53 51 AM

micahstairs (approved):
Great work!!

PR #464, opened 2022-05-16
Remove padding from age squares

micahstairs:
I removed the padding on each side of the age squares. This padding was originally only for the age 10 & 11 icons, I believe, but now we use the same CSS for all ages.

I made more substantial changes to the age squares in the game log (.N class).

Before: 
Screen Shot 2022-05-16 at 10 14 40 AM

After:
Screen Shot 2022-05-16 at 10 18 32 AM

### DIFF
--- a/innovation.css
+++ b/innovation.css
@@ -1031,24 +1031,22 @@
 .square.basic.icon_action {
   background-position: -48px 0px;
 }
-.square.N {
-  width: 14px;
-  height: 14px;
-}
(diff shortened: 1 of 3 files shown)


lordmcfuzz:
So what does the age 10 box look like?

micahstairs:
Screen Shot 2022-05-17 at 6 29 27 AM

It still looks good in my opinion!

lordmcfuzz (approved):

Issue #448, opened 2022-05-13
Transferring from forecast doesn't seem to work properly


ultimatefiend:
I attempted this via Almanac and it doesn't do the transfer:

           // First reveal the card to prove that it has a bonus.
            $options = array(
                'player_id' => $player_id,
                'n' => 1,
                'can_pass' => true,
                
                'owner_from' => $player_id,
                'location_from' => 'forecast',
                'owner_to' => $player_id,
                'location_to' => 'revealed',
                
                'with_bonus' => true,
            );
            break;

and it does this:

![image](https://user-images.githubusercontent.com/11416698/168287338-e11adabb-f936-4dd6-a885-318726a4374d.png)

There is a card in the forecast with a bonus when this happens.

micahstairs:
I found a bug in your PR in encodeLocation() which may or may not fix this, I'm not sure.

ultimatefiend:
this should be fixed in the age 4 PR i just submitted.

PR #470, opened 2022-05-21
Kahlia echoes age4

ultimatefiend:
All of the Age 4 cards are here.

The only problematic one is Barometer which requires a blue to be revealed while it is being returned.  I couldn't quite get that right.  The user should be prompted to reveal the blue card to prove their worthiness for Destiny, but I couldn't think of a good way to do that while the cards were being returned.

This fixes the bug in Almanac as well and some fixes for the card descriptions.


ultimatefiend:
#448

micahstairs:
For Barometer, here's an idea... If Destiny is still claimable and the player has at least one blue card, before getting them to return all of the cards from the forecast, first prompt them to reveal and return one blue card.

ultimatefiend:
I thought about that, but then the player has to return the blue card first (which may not be desirable).

It could be a multi-level selection where they first reveal a blue card from their forecast (and then put it back) and then return all of the cards in their forecast.  If the first part succeeds, they would get Destiny.  It sounded really complicated, but might be a better option.

micahstairs:
I like your suggestion! I'll let you implement that before I review your changes

ultimatefiend:
Yeah, Barometer is a weird one. I'm not sure how you do it without revealing whether the player has a blue in their forecast when they don't want to go for Destiny.

micahstairs:
With the wording on Destiny, I don't think players are allowed to not go for it?

ultimatefiend:
You may return all of cards so they could elect to not go for it if they are about to get it anyway.

The problem is that you have to ask to reveal the blue cards before they are returned.  I guess they could just pass on the reveal of the blue card....maybe that will be sufficient?   They could theoretically still return their cards from their forecast at that point but I don't know why you would.

ultimatefiend:
Some rewording on this text would really be helpful like :

"You may reveal and return a blue card from your forecast.  If you do, claim the Destiny achievement and return all cards from your forecast."

micahstairs:
Ah yes, that would be more clear. I think proceeding with that as our
implementation is reasonable! Even though we won't actually be changing the
card's text.

ultimatefiend:
Could you add it to the list of things to talk about in your next email to Carl?  I'd like to get his ok on the interpretation/implementation

micahstairs:
Sure I'll ask him about it. I guess we can leave your implementation as-is for now and we can revisit this later.

Ready for me to review this?

ultimatefiend:
No, let me do it as described above first.  RIght now it fails to run.

micahstairs:
By above, I'll assume you mean: "You may reveal and return a blue card from your forecast. If you do, claim the Destiny achievement and return all cards from your forecast."

ultimatefiend:
yes

ultimatefiend:
ok good to review now

Issue #225, opened 2022-02-05
Mona Lisa requires a non-negative integer selection


ultimatefiend:
**NOTE:** Before you begin working on this, please **assign this issue to yourself**. If you do not do this, then there's the possibility that 2 people will end up doing the same task, which is not a good use of time.

Mona Lisa (#104) requires that any number be selected (not an age number).  In real terms, the range needed can be from 0-106 (all yellow cards in hand from all expansions including the relic).

micahstairs:
Aiae suggested:

> It asked me the number and then color, but could it be possible to ask for both number and colour at the same time, with a "Confirm" button at the bottom once you're happy with your choices

micahstairs:
Given that we already have to make a custom chooser for this card, I think this is a great idea!

micahstairs:
After thinking about this a bit, I'm wondering if we can change how we phrase the number selection. Instead of getting them to choose the total number of that color that will end up in their hand, would it work if we ask them to guess how many of a certain color they are going to draw? Then we only need to give them the option to select 0-5, and we wouldn't need to add a new UI element to handle this.

micahstairs:
We are concerned the above will lead to confusion.

Here's a mock of the initially proposed UI element:

<img width="1428" alt="Screen Shot 2022-03-11 at 10 38 19 AM" src="https://user-images.githubusercontent.com/7231485/157899235-864a65ba-a6e2-4598-95ce-22511b218b2c.png">

micahstairs:
Actually eoc has a better idea:

"In a drop-down, you can choose your desired color, and then six buttons appear: one for each possible amount of cards that color in your hand after resolving the draw. So if you already had 4 green cards and selected Green, the option buttons would be: 4/5/6/7/8/9 (and they would represent the new amount of green cards, not the ones drawn)"

micahstairs:
Actually I'm not a fan of making players use a dropdown. The latest proposal is to do something like this (except with 2 extra buttons on it):

<img width="372" alt="Screen_Shot_2022-03-12_at_8 25 11_PM" src="https://user-images.githubusercontent.com/7231485/159134627-2dd672a8-94bd-4ff4-a47e-7124b8d70da3.png">

micahstairs:
Marking this as a bug since 0 is not currently selectable.

PR #471, opened 2022-05-21
Add non-negative integer selection to Mona Lisa (#225)

micahstairs:
It defaults to however many cards you currently have of that color. I display 6 options since that allows them to pick X ... X+5 (since Mona Lisa draws 5 cards). If you click the arrows it will change the selection of numbers that you have. In practice, users will never need to click these button, but this gives them the flexibility to pick different numbers if they really want to.

Screen Shot 2022-05-21 at 7 42 26 AM
Screen Shot 2022-05-21 at 7 42 38 AM
Screen Shot 2022-05-21 at 7 42 46 AM


ultimatefiend:
That's an elegant change!  I like it.

ultimatefiend (approved):

Issue #457, opened 2022-05-14
Audit remaining usages of loadPlayersBasicInfos


micahstairs:
We need to make sure any usage of loadPlayersBasicInfos is compatible with players being eliminated.

ultimatefiend:
I did a check of this when I was working on Exxon but a second look wouldn't hurt.

PR #476, opened 2022-05-23
Clean up usage of loadPlayersBasicInfos (#457)

micahstairs:
I've gone through all callsites of loadPlayersBasicInfos and removed the ones where we technically want to ignore eliminated players. 

I've manually tested all of the affected cards:  Clothing, Construction, Self Service, A. I., Maastricht Treaty, Socialism, and Specialization.


ultimatefiend (approved):
I like the simplification of robotics/software detection for AI.  It looks good.

Issue #451, opened 2022-05-14
Add log message to explain digging artifacts when the base pile is empty


micahstairs:
Many players get confused when they (or an opponent) gets an artifact of a higher age as a result of base piles being empty. We should add a quick log message when this happens to avoid this confusion.

micahstairs:
I played around with this and wasn't happy with the wording of any of the log messages I came up with. I'm considering a different idea now: let's reduce the opacity of expansion decks if the accompanying supply pile is empty. This will make it easy to tell which decks are currently inaccessible.

PR #477, opened 2022-05-23
Make expansion decks transparent if their base deck is empty (#451)

micahstairs:
What do you think of this feature? It should also make it more obvious which artifact will go on display since it's easier to tell which ages are skipped.

Screen Shot 2022-05-23 at 12 59 51 PM


ultimatefiend:
I like that idea.  that makes it very obvious.

Issue #463, opened 2022-05-14
Make it possible to read relic's text at all times


micahstairs:
We received feedback about this during the alpha: https://boardgamearena.com/bug?id=63670.

I agree that it should be possible to hover over the relic cards and read its text, regardless of where the relic is on the board.

We will choose not to let players read relics if the associated set is not in play though, since that means that it will never end up being executed.

PR #492, opened 2022-06-07
Add tooltip to relic regardless of location (#463)

micahstairs:
Screen Shot 2022-06-07 at 7 57 45 PM


ultimatefiend (commented):
Is there logic in here that prevents relics that should not be visible (i.e. safety pin) from being visible via tooltip?

micahstairs:
> Is there logic in here that prevents relics that should not be visible (i.e. safety pin) from being visible via tooltip?

That's what the canShowCardTooltip function does.

ultimatefiend (approved):

Issue #494, opened 2022-06-08
Add game preference to use simplified graphics


micahstairs:
I've received requests from multiple players that it's hard for them to read the numbers on the back's of cards. I don't personally have any issues reading it, but I can understand why this might cause accessibility issues for certain players.

ultimatefiend:
is this done now?

micahstairs:
Yup!

PR #495, opened 2022-06-09
Add game preference to use simplified graphics (#494)

micahstairs:
Screen Shot 2022-06-08 at 8 44 51 AM

### DIFF
--- a/gameoptions.inc.php
+++ b/gameoptions.inc.php
@@ -169,5 +169,14 @@ $game_preferences = array(
             4 => array('name' => totranslate('Enabled (long timer)')),
         ),
         'default' => 1
+    ),
+    110 => array(
+        'name' => totranslate('Simplified graphics'),
+        'needReload' => true,
+        'values' => array(
+            1 => array('name' => totranslate('Disabled')),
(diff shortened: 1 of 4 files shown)


ultimatefiend (commented):
To distinguish between the number of cards and the age of the cards, could we add a "x" in from the of the number for the card count?  (or something else to make it obvious)

micahstairs:
If we are going to make a change like that, should we look at making each column of cards take up less vertical space? I'm thinking ahead to when we have 5 different sets of cards being used at once. If things were condensed enough that we could have 5 columns side by side, that would make it easier to look at the decks and know what's going on.

micahstairs:
I tried adding an `x` after the numbers, however, I wasn't able to do so because we are using BGA's counter UI component there and I think it's expecting to be working with integers.

micahstairs:
@ultimatefiend  I understand there is still room for improvement here, but do you think this is more accessible to players than it was before this change? In other words, do you think we need to improve this further before we release or is it okay for now?

ultimatefiend:
Yes, it is fine for now.

PR #504, opened 2022-06-18
Kahlia echoes age8 2

ultimatefiend:
Implement most of the 8s, and fix some 7 issues.


micahstairs (approved):
Pretty good! I did see one major bug in the Bandage implementation and left a TODO for that.

Issue #501, opened 2022-06-12
Implement Age 9 cards for Echoes Expansion


ultimatefiend:
This issue will track the implementation of all cards from this age for the Echoes expansion. Please update this comment before working on any of these cards (if the work is unblocked) so that we don't have duplicate work being done.

  - [x] **415- Calculator**  (BLOCKED)
  - #497
  - [x] **416- Laser** (BLOCKED)
  - Not sure what a "standard" achievement is.  Does this include the first 9 achievements or does it also include other achievements that were added (i.e. toothbrush or chopsticks)?
  - [x] **417- Helicopter** (BLOCKED)
  - #503
  - [x] **418- Jet** 
  - [x] **419- Credit Card** 
  - [x] **420- Email** 
  - [x] **421- ATM** 
  - [x] **422- Wristwatch** 
  - [x] **423- Karaoke** (BLOCKED)
  - #497
  - [x] **424- Rock**

PR #506, opened 2022-06-25
#501 - all but 3 (and half karaoke)

ultimatefiend:
I didn't do the first 3 due to some missing features.

Karaoke is partially done.  I got halfway through before realizing I couldn't select bottom cards...heh.


micahstairs (approved):
Great work! I just made a few small changes.

PR #507, opened 2022-06-26
Make it possible for players to tell which type each card is

micahstairs:
This was requested in https://boardgamearena.com/bug?id=66491.

Screen Shot 2022-06-26 at 9 52 19 AM
Screen Shot 2022-06-26 at 9 57 27 AM

### DIFF
--- a/innovation.css
+++ b/innovation.css
@@ -746,6 +746,12 @@
   right: 22px;
 }
 
+.card_type {
+  padding-top: 10px;
+  font-style: italic;
+  text-align: center;
+}
+
 .card_title {
   color: white;
   font-family: Tahoma, Arial, sans-serif;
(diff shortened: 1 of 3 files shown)


ultimatefiend (changes requested):
I'd change the if else to a switch so it will be easier to maintain later.

Otherwise, it looks great.  I like the positioning of the text.

micahstairs:
I agree that using a switch statement makes sense once the other expansions are added. I'll defer that change to the main-echoes branch, but I'm not in a rush to make that change in main-echoes either, because minimizing merge conflicts is helpful whenever I need to rebase main-echoes onto main-dev.

ultimatefiend:
ok cool.  Don't forget about pushing it to main-echoes though. :)  Its a nice feature.

Would you add it after the alpha is released into production?

micahstairs:
This will be included in the next push to alpha (and will be rebased into main-echoes next time I do a rebase). What I meant was that I'm not going to add the switch statement until later!

PR #508, opened 2022-06-26
Fix bug which sometimes resulted in wrong deck opacities

micahstairs:
I was able to replicate this bug reliably when the last card of a deck was going to a player's own score pile.


ultimatefiend (approved):
Seems good.  The delay should help and I like the comment about possibly finding a better solution in the future.

micahstairs:
Yeah and the delay feels pretty natural actually.

Issue #462, opened 2022-05-14
Clear up confusion about when special achievements are awarded


micahstairs:
Many players have voiced confusion (https://boardgamearena.com/bug?id=62821) about when special achievements are awarded (as a result of us changing the timing of the special achievement check, after consultation with Chris). 

I'm in the process of emailing back and forth with Chris to reach a resolution on this.

We definitely need to get this story straight before we can release the expansion.

micahstairs:
Marking as blocked until I get a response back from Chris.

micahstairs:
I just sent another email. I'm trying to reach a resolution soon on this.

micahstairs:
Finally got a response. He agreed with my proposal:

```
- Remove the word "immediately" from all of the base special achievements except for Monument. Similar to the special achievements in Cities, it will confuse players more if Monument isn't awarded the instant they finish scoring or tucking 6 cards.
- Keep the timing of the other special achievements to check at the end of the effect
```

ultimatefiend:
so does that mean Glory and Victory are awarded immediately or not?  I'm not clear on that.

micahstairs:
I think we want to award those immediately! We only want to wait for achievements which depend on examining the board state.

ultimatefiend:
Those will be easy then.  I like it.

PR #512, opened 2022-07-01
Revise special achievement timing logic (#462)

micahstairs:


innovation.game.php
```diff
@@ -2833,7 +2840,12 @@ function getActivePlayerIdsInTurnOrderStartingWithCurrentPlayer() {
             $current_player_id = self::getCurrentPlayerUnderDogmaEffect();
         }
 
-        $current_player_no =  self::getUniqueValueFromDB(self::format("SELECT player_no FROM player WHERE player_id={current_player_id}", array('current_player_id' => $current_player_id)));
+        if ($current_player_id < 0) {
```
ultimatefiend, line 2843: what is this for?

micahstairs: This is necessary because we re-added the special achievement check for Monument after each card transfer, and when it's the initial meld, it's technically everyone's turn so we don't have an active player. So we arbitrarily choose a player (in this situation no one would get a special achievement on an initial meld anyway, but maybe in debug mode, we'd trigger it by melding the right cards)

ultimatefiend (approved):

Issue #449, opened 2022-05-14
Sunglasses requires a very complicated choice


ultimatefiend:
The text on Sunglasses is the following:

You may either splay your purple cards in the direction one of your other cards is splayed, or you may splay one of your other colors in the direction that your purple cards are splayed.

Those this could present a quite varied number of options to the user depending on the state of their board.

For example, if purple were splayed up and the other piles splayed left and right and one pile was unsplayed, then there would be the following options:

splay blue up, splay yellow up, splay red up, splay green up, splay purple right, splay purple left, unsplay purple

I don't know how to set this up with the current interaction setup.  I think it would require something new, but I could be wrong.  Telegraph (age 7) also requires this splay imitation which appears to be unique to Echoes.

micahstairs:
What if you break it into two steps? You could get them to pick between the two options first, and then you could get them to pick the color?

ultimatefiend:
I guess that would work and would simplify the logic.  Should we have a way to jump back to the first step in case they have second thoughts?  I guess we don't do that with other cards and we have the confirmation feature now.

micahstairs:
The confirmation feature currently only supports the initial click on a card to meld/dogma.

However, for simplicity, I propose we don't give them a way to jump back to the first step. And then we could figure out something later if it is a pain point.

ultimatefiend:
is this done now?

micahstairs:
There's a few small TODOs. But I think it's fine to close.

PR #513, opened 2022-07-02
#449 - sunglasses

ultimatefiend:


micahstairs (approved):
I made some minor changes, but in general I think this is the right approach. I left some TODOs because my interpretation of the card is that the player can choose to do neither of the effects. I could be wrong though.

Issue #514, opened 2022-07-03
Add table-based auxiliary value in order to store long lists of integers


micahstairs:
This will be used by cards like Lever. This will also be used by cards like Radio Telescope once we add some sort of option to tell the interaction to select all cards IDs in the table.

ultimatefiend:
I see.  I can see two possible interactions that would be useful:

1.  Allow a user to select all of the card ids in the list regardless of where they are.  This could be helpful for Hot Air Balloon. #489
2. Allow repeat age selections (crossword/lever).  This would make that interaction more natural.

micahstairs:
The only possible outstanding piece of work is the part you had about "Allow repeat age selections (crossword/lever). This would make that interaction more natural."

However, I'll close this, and we can file a separate issue for that if we want to make that improvement.

PR #515, opened 2022-07-03
Add table-based auxiliary values (#514)

micahstairs:


ultimatefiend (commented):
A few comments:

1.  Its not necessarily specific to card ids right?  (the comments suggest otherwise) It looks like it could be a variable length array of any integer.
2. How do you detect the length of the stored array?  Do you need to extract the whole array and then do a "count"?
3. Will this work with nested execution?  It looks like it is intended to do that, but wanted to make sure.

micahstairs:
> A few comments:
> 
> 1. Its not necessarily specific to card ids right?  (the comments suggest otherwise) It looks like it could be a variable length array of any integer.
> 2. How do you detect the length of the stored array?  Do you need to extract the whole array and then do a "count"?
> 3. Will this work with nested execution?  It looks like it is intended to do that, but wanted to make sure.

1. Correct.
2. The length is stored in the 0th index.
3. Yup! It will work for nested execution.

ultimatefiend (approved):

Issue #372, opened 2022-04-11
Implement Cities special achievements


ultimatefiend:
All of the special achievements occur in either on meld event (the splaying achievements) or during a tuck dogma/endorse action (flag, fountain).
 The splaying achievements should be pretty simple.  The tucking ones shouldn't be too hard.

ultimatefiend:
There are some special icons on the special achievement cards (left, right, up arrow, flag and fountains)

ultimatefiend:
@micahstairs This will require additions to the database.  We basically need to set flags when the conditions for the special achievements are met (sort of like counters for monument).  How big a deal is it?

ultimatefiend:
or maybe this could be solved with a global variable?  I'm not sure why the monument trackers are stored in the DB

micahstairs:
I haven't looked closely at the Cities special achievements and thought about the best way to implement them yet 

But to answer your questions, global variables are still stored in the DB. Also the scoring and tucking counters are in the players table because each player needs its own counter (since you can get the special achievement even on an opponent's turn).

ultimatefiend:
All of the Cities special achievements are like Monument in that they are earned through actions during a turn and are not dependent on the final state of the board after an action has occurred (like the other Base and all Echoes specials).

3 of them are earned by melding a card onto a pile of a matching splay (left, right, up) and 2 of them are earned by tucking a city card with a flag or fountain.  So, I think you need to set a flag in the event the condition is met during the turn (like Monument).  It doesn't need to be a count like Monument.

Also, with the transition from awarding them immediately to awarding them at the end of an action, the flag would make sense.

I guess we could also keep the immediate aspect of these specials and that would prevent this complication.

micahstairs:
I was going to email Chris this week and get his thoughts on the "immediately" aspect of this.

ultimatefiend:
ah ok.  sounds good.

micahstairs:
Actually wait, we can definitely do them immediately here since they happen as a result of a Meld action. We don't need to worry about the waiting until the end of the current dogma effect in this case. And because we can do it immediately, we don't need to worry about storing anything in the DB.

ultimatefiend:
ok, excellent.  I will do it that way then.

micahstairs:
For the two that are a result of tucking, that will require a monument-like behavior since that happens during a dogma effect.

ultimatefiend:
[...]

PR #517, opened 2022-07-05
#372 - victory and glory

ultimatefiend:


micahstairs (approved):
Looks good! I love how simple this is.

Issue #436, opened 2022-05-07
Implement Lever's special requirement for tracking pairs of matching cards returned


ultimatefiend:
This is the text on the card:

You may return any number of cards from your hand. For every two cards of matching value returned, draw a card of value one higher.

The "return any number of cards" part of this card means that you can return multiple of the same age to draw up to the next age.  For example, if you have 4 2's in hand, you can return all 4 of them to draw 2 3's.  So tracking how many of each age returned is important.

So figuring out how to store both ages and counts is the challenge here.  This card is kind of like currency, but the number of each age returned is important.

ultimatefiend:
A 'multiselect' feature might be useful here (like we discussed with Tools briefly).

ultimatefiend:
this can be closed I think.

PR #518, opened 2022-07-07
#436 - lever!

ultimatefiend:


micahstairs (approved):
I think there's a few improvements we could make to this card, but it seems to be working fine for now!

Issue #502, opened 2022-06-12
Implement Age 10 cards for Echoes Expansion


ultimatefiend:
This issue will track the implementation of all cards from this age for the Echoes expansion. Please update this comment before working on any of these cards (if the work is unblocked) so that we don't have duplicate work being done.

  - [x] #509 (BLOCKED)
  - [x] **426- Human Genome** (BLOCKED)
   - #497
  - [x] **427- Camcorder** 
  - [x] #510 (BLOCKED)
  - [x] **429- GPS** 
  - [x] **430- Flash Drive** 
  - [x] **431- Cell Phone** 
  - [x] **432- MP3** 
  - [x] **433- Puzzle Cube** (BLOCKED)
  - #497
  - [x] **434- Sudoku**

PR #519, opened 2022-07-12
#502 - cell phone, camcorder, gps, flash drive, mp3

ultimatefiend:


micahstairs (approved):
Looks great! I only made a few minor tweaks.

PR #524, opened 2022-07-14
add note about php needing to be v7

Nathansbud:


micahstairs (commented):
Actually this needs to be based off of main-dev, not main-echoes, or else it won't appear in the documentation everyone looks at (https://github.com/micahstairs/bga-innovation).

Nathansbud:
Ah, my bad! I didn't want to merge to main-dev, as you mentioned it was the production branch (I'll remake the PR). How do builds get out to BGA?

Issue #523, opened 2022-07-14
Skip the Bicycle interaction if no cards in hand and score pile


ultimatefiend:
https://boardgamearena.com/bug?id=63591

This should be very easy to implement.

micahstairs:
Just deployed this to production.

PR #526, opened 2022-07-15
#523 - bicycle

ultimatefiend:


lordmcfuzz (approved):
Looks good to me.

The comment rider probably should be in the else block, as it kinda reads like it is for the if block

micahstairs (commented):
Thanks for implementing this but you did it on the wrong branch. I copied your code to main-dev in https://github.com/micahstairs/bga-innovation/commit/c8dee114147dfe5c309a935839db77a33408feb0.

PR #527, opened 2022-07-16
#504 implement crossword

Nathansbud:
Took a crack at doing Crosswords. Logic somewhat similar to Lever although it has a slightly janky choice interface. Might be worth somebody more experienced making it prettier, I wasn't sure how to nicely represent that you had, say, 2 1s left to draw, 1 8 left, ...functional, but perhaps confusing?


innovation.game.php
```diff
@@ -13686,7 +13690,24 @@ function stPlayerInvolvedTurn() {
...
+                $vis_bonuses = self::getVisibleBonusesOnBoard($player_id);
+                $num_bonuses = count($vis_bonuses);
+                // if no bonuses visible, nothing happens
+                if($num_bonuses > 0) {
+                    $bonus_counts = array(0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0);
+                    foreach ($vis_bonuses as $bonus) {
+                        $bonus_counts[$bonus - 1]++;
+                    }
+                    
+                    $step = 1;
```
ultimatefiend, line 13706: This is not necessary.  The step will be one if step_max is defined.

innovation.game.php
```diff
@@ -13686,7 +13690,24 @@ function stPlayerInvolvedTurn() {
                 self::executeDraw($player_id, 7, 'score');
                 self::executeDraw($player_id, 7, 'hand');
                 break;
-
+            
+            // id 413, Echoes age 8: Crossword
+            case "413N1":
```
ultimatefiend, line 13695: You could draw a card automatically if only one bonus present (presumably the one on Crossword)

innovation.game.php
```diff
@@ -22829,7 +22869,16 @@ function stInterInteractionStep() {
...
                     self::executeDraw($player_id, $value_to_score, 'score');
                     break;     
-                
+                // id 413, Echoes age 8: Crossword
+                case "413N1A":
+                    if(array_sum(self::getAuxiliaryArray()) > 0) {
+                        $step--;
+                        self::incrementStep(-1);
+                    } else {
+                        $step++;
```
ultimatefiend, line 22878: What step are you skipping here?

Only the most complicated state machines within a card would require skipping a step.

ultimatefiend (changes requested):
See comments.  Looks pretty good overall.

I presume you have tested this?

innovation.game.php
Nathansbud: Maybe I am misunderstanding how $step works entirely; I was trying to keep all logic happening in "431N1A" steps, as the choice selection needs to be repeated N times until all N cards have been drawn (as the player controls the order). So, decrease the $step to keep things at 0, and increase step to finally end the dogma.

That said, I was iffy on the difference between $step-- and incrementStep(-1). Mostly, I was trying to copy the logic of other repeated choice cards I saw—Lever (which kahlia mentioned basing this off), Oars.

Nathansbud:
Went back in and added comments to resolve auto-resolving. 

One bug right now: 11 can't be displayed as an age marker in chooseValue, which means all values will be chosen, then things will auto-end with the 11. Either this can be merged in now, and this bug can be resolved later once this logic is added,  or the MR can stay open for a bit until this is added.

micahstairs:
Feel free to add TODO(ECHOES) for the outstanding work item.

Also, it's nice to automate choices, but sometimes it can add a lot of complexity to cards. As a general rule, it's better to generalize the automation as much as possible. For example, it would be better if we added generic logic which automatically chooses a value if there's only one value to choose from (similar to how we introduced auto-selection for clicking on cards). What do you think?

micahstairs:
Also we don't want to add that generic automation to main-echoes. It would be better to add the generic automation to main-dev so that we can release it publicly on the earlier side.

innovation.game.php
```diff
@@ -22829,7 +22886,28 @@ function stInterInteractionStep() {
...
+                        // if only one value remains, find the value and draw that many cards
+                        for($i = 0; i < $bonus_counts; $i++) {
+                            if($bonus_counts[$i] > 0) {
+                                for($d = 0; $d < $bonus_counts[$i]; $d++) {
+                                    self::executeDraw($player_id, $i + 1, 'hand');
+                                }
+                                break;
+                            }
+                        }
+                        $step++;
```
ultimatefiend, line 22907: You don't need to specifically advance the step as that happens automatically.  You can think of the incrementStep function as your only way to keep the step from advancing.

In rare occurrences, you might need to skip an interaction, but I don't think its necessary here.

innovation.game.php
```diff
@@ -22829,7 +22886,28 @@ function stInterInteractionStep() {
...
                     self::executeDraw($player_id, $value_to_score, 'score');
                     break;     
-                
+                // id 413, Echoes age 8: Crossword
+                case "413N1A":
+                    $bonus_counts = self::getAuxiliaryArray();
+                    $remaining_values = count(array_filter($bonus_counts));
+                    // if more than 1 value remains, repeat the choice loop
+                    if($remaining_values > 1) {
+                        $step--;
```
ultimatefiend, line 22895: This part looks good.  You need to decrement the step to return to the current interaction.

innovation.game.php
```diff
@@ -22829,7 +22886,28 @@ function stInterInteractionStep() {
...
+                        self::incrementStep(-1);
+                    } else {
+                        // if only one value remains, find the value and draw that many cards
+                        for($i = 0; i < $bonus_counts; $i++) {
```
ultimatefiend, line 22899: This is a good optimization.  If they are all the same value, then there is no choice to be made.

ultimatefiend (changes requested):

ultimatefiend:
Is this still being worked on?  Should someone else finish this up?

micahstairs:
These changes were copied into https://github.com/micahstairs/bga-innovation/pull/653 since we haven't heard from Nathan in a long time

Issue #530, opened 2022-07-16
Make green and blue backgrounds easier to distinguish 


micahstairs:
One of the biggest complaints is that the 3rd edition blue and green card backgrounds are too similar.

micahstairs:
Be sure to update the Cities card backgrounds for consistency. Those files are already in the main-dev branch.

micahstairs:
I pushed this to production!

PR #531, opened 2022-07-16
Tint green cards to distinguish blue and green more easily

normaljosh:
See issue #530 

Tinted card backgrounds, hexes, and cities special icons to make greens more intense using imagemagick. Tinting blues made the shadows look off, so I just adjusted greens.


micahstairs (approved):
Thanks!

Issue #536, opened 2022-07-28
Add a comment to the log when Baghdad Battery does not score any cards


ultimatefiend:
https://boardgamearena.com/bug?id=68655

This is a good suggestion.  We can add a log message that mentions that the cards were from the same expansion:

"The melded cards are from the same card set, so no cards are scored."

Issue #534, opened 2022-07-27
Some cards when executed with an "n" of 0 do not execute as expected


ultimatefiend:
This was brought up via a bug report:

https://boardgamearena.com/bug?id=68739

When executing Gujin, you can execute a card on someone else's board as if its on your own board...which means you can fewer than 2 leaves for Reformation.  This should lead to no cards being tucked, but right now it doesn't.

There are some other cards that would have similar behavior.  The safest solution is to ignore any interaction that specifies an "n" of 0 during the interaction specification phase.  This would allow for some simplification of other cards as well.

PR #544, opened 2022-08-02
#536, #534

ultimatefiend:
add comments to baghdad and also detect an n = 0


micahstairs (commented):
You had this on main-echoes instead of main-dev, so I cherrypicked it, made minor changes, re-tested it and deployed it.

PR #553, opened 2022-08-27
Add promote mechanism

micahstairs:
There are a few improvements we can make to the user experience, however, it seems to be quite functional!

I had to make some changes to the state machine, but I believe all of the changes can be safely rolled out to ongoing games when it comes time to release.


innovation.game.php
```diff
@@ -120,6 +120,7 @@ function __construct()
             'with_bonus' => 87,
             'without_bonus' => 88,
             
+            'melded_card_id' => 95, // ID of the card which was melded
```
lordmcfuzz, line 123: was this meant to double up with relic_id?

micahstairs: I did consider re-using that global, but I thought I'd better use a properly named variable.

Also, just now I realized that there are moments in the same where we need values in both at the same time (it's possible to meld, triggering both an artifact/relic as well as a promoted card). If we reused the relic_id then the originally melded card would be lost by the time we got to choose a foreshadowed card.

micahstairs: Oh now I see I accidentally used the same index for both! Good catch, I will need to fix that.

micahstairs: Fixed!

ultimatefiend (approved):

PR #557, opened 2022-08-27
Automate Explosives if there are 3 cards or less in hand

micahstairs:
This card wasn't possible to fully automate using a general approach, so I needed to add card-specific automation to it.

I manually tested cases with 0, 1, 2, 3, and 4 cards in hand to make sure everything was working correctly.

Also, due to the nature of this change, it won't break any ongoing games when this is deployed.

### DIFF
--- a/innovation.game.php
+++ b/innovation.game.php
@@ -9474,28 +9474,30 @@ function getOwnersOfTopCardWithColorAndAge($color, $age) {
             // id 68, age 7: Explosives
             case "68D1":
 
-                // Automate selection if the player has no more than 3 cards in hand
-                $cards = self::getCardsInLocation($player_id, 'hand');
-                if (count($cards) >= 1 && count($cards) <= 3) {
-                    foreach ($cards as $card) {
-                        $card = self::getCardInfo($card['id']);
-                        self::transferCardFromTo($card, $launcher_id, 'hand');
+                // Automate taking as many highest cards as possible
+                $num_cards_in_hand = self::countCardsInLocation($player_id, 'hand');
+                $cards_by_age = self::getCardsInLocationKeyedByAge($player_id, 'hand');
+                $num_cards_left_to_transfer = 3;
+                for ($age = 10; $age >= 1; $age--) {
+                    if (count($cards_by_age[$age]) <= $num_cards_left_to_transfer) {
+                        foreach ($cards_by_age[$age] as $card) {
+                            $card = self::getCardInfo($card['id']);
+                            self::transferCardFromTo($card, $launcher_id, 'hand');
+                            $num_cards_left_to_transfer--;
+                            $num_cards_in_hand--;
+                        }
+                    } else {
+                        break;
                     }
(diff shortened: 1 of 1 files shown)


ultimatefiend (changes requested):
Hmm...that is one weird card implementation.  I had to look at the rest of the implementation to make sense of these changes.

This change does not automate when there are three clear highest cards and they have the same value and other cards are also in hand.  For example, 3 7's and a 2 in hand.  The player would need to select each card individually (except for the last one).

Also, if there are zero cards in hand why is step_max set to anything?  Shouldn't the dogma just end at that point?  Why would it continue through to the other steps?  If it did end, you would know whether a card was transferred and the aux value flag wouldn't be needed either (you could look at the number of cards in hand at that point).  You could also say that if someone makes it to the interaction portion they must have more than 3 cards in hand and won't do the draw.

* after rereading *
I see how the break's are being used here and I find that confusing.  I'd prefer that the code not use breaks (unless its a loop) otherwise following the logic becomes more confusing.

micahstairs:
`This change does not automate when there are three clear highest cards and they have the same value and other cards are also in hand. For example, 3 7's and a 2 in hand. The player would need to select each card individually (except for the last one).`

If there are three clear highest cards then this will be handled by the generalized logic for automatically selecting multiple cards: https://github.com/micahstairs/bga-innovation/pull/558.

`Also, if there are zero cards in hand why is step_max set to anything? Shouldn't the dogma just end at that point? Why would it continue through to the other steps? If it did end, you would know whether a card was transferred and the aux value flag wouldn't be needed either (you could look at the number of cards in hand at that point). You could also say that if someone makes it to the interaction portion they must have more than 3 cards in hand and won't do the draw.`

I did also consider this! The reason I let it continue to the third step is so that we get a log message saying "No card matches the criteria of the effect". Maybe we should decide if we want to go out of our way to show that message. The code will be simpler if we don't show it.

`I see how the break's are being used here and I find that confusing. I'd prefer that the code not use breaks (unless its a loop) otherwise following the logic becomes more confusing.`

Alright, I can revise this to use if/else statements once we decide our plan of action for the zero card situation.

ultimatefiend:
I would say the code in the switch statement should prevent the start of the interaction that won't do anything.  It could print a log message for that case that would state that no cards were available to transfer or something like that.  I think that's better than jumping to the last step that is guaranteed to not do anything.

micahstairs:
Okay I agree with you. I'll simplify it. And I found other cards that don't log anything in situations like this where it's obvious that no card was eligible to be selected.

micahstairs:
Alright, I've revised the automation. I think it's more in line with what you were initially expecting (it also automates a few special cases that wasn't being automated before).

I can't remove the usage of the auxiliary value yet or it could break ongoing games, but we can remove that later. I've added TODOs so that we don't forget to clean it up.

ultimatefiend (approved):
This looks good.  Its a slick implementation.

PR #560, opened 2022-08-28
Drop caps on 'SINGLE' since it was too strong

micahstairs:


ultimatefiend (approved):
I would change it to say something like :

"a single action on their first turn"

Not required, but might make it clear the single action is only on this current turn and doesn't affect subsequent turns.

PR #567, opened 2022-09-03
Add custom warning to Tools

micahstairs:
One of the biggest requests from players is to improve the UX with Tools (https://boardgamearena.com/bug?id=18164). this is my proposal for addressing it.
Screen Shot 2022-09-03 at 11 51 27 AM

Please note that I structured the JS code in a way that it will make it easy to add confirmation buttons on a timer like we do for meld/dogma actions. I can easily see us wanting to give players an option to confirm card selections as well.

### DIFF
--- a/innovation.game.php
+++ b/innovation.game.php
@@ -7982,6 +7982,8 @@ function getOwnersOfTopCardWithColorAndAge($color, $age) {
             'splay_direction' => $splay_direction,
             'splay_direction_in_clear' => $splay_direction_in_clear,
             'color_pile' => $splay_direction === null && $location_from == 'pile' ? self::getGameStateValueAsArray('color_array')[0] : null,
+            'card_interaction' => $code,
+            'num_cards_already_chosen' => $n,
             
             // Private info
             '_private' => array(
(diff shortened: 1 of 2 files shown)


lordmcfuzz (approved):

ultimatefiend (commented):
The confirm of the "3" specifically is a good thing to do.  Does this prompt on all three returns or just the first?  

It seems a bit wordy.  Do we really need the second part of the explanation?  Alternatively, it could state what won't happen (they won't draw 3 ones if the 3 is returned in the first dogma effect).  That would make the reason for the prompt clearer.

micahstairs:
It only prompts on the return of the first card (if it's a 3). This is why the backend is passing that `num_cards_already_chosen` variable now, so that we can tell if it's the first card being returned or not.

I agree that it's pretty wordy. I'm definitely open to suggestions on new wordings!

micahstairs:
How about `Are you sure you want to return a 3 as part of the first non-demand effect? This won't allow you to draw three cards.`?

micahstairs:
This is what it looks like now:
Screen Shot 2022-09-08 at 8 49 17 PM

micahstairs:
I shortened the warning message even more:

Screen Shot 2022-09-08 at 8 54 24 PM

Issue #377, opened 2022-04-14
Add Echoes hexagon icons to spritesheets


ultimatefiend:
We need the echoes hexagons to implement the expansion.

micahstairs:
We will put these right below the Artifact icons, following the same pattern.

micahstairs:
Actually that spritesheet is already really big and close to our limit, so we should create a new spritesheet instead of adding it to the existing one.

To make the spritesheet as small as possible, I propose that we make it 11 columns and 10 rows. There are 15 age 1 cards, so we will put the 11th-15th icons in the 11th column.

PR #573, opened 2022-09-04
make spritesheet for echoes hex icons

normaljosh:
Creating separate spritesheet for echoes hexagons as described in #377


micahstairs (approved):
Thanks for doing this! The only icon that's wrong is the Helicopter icon (red age 9). I can fix it in a follow-up PR. I'll merge this!

Issue #503, opened 2022-06-15
Radio Telescope needs a way to select more cards


ultimatefiend:
Radio Telescope (age 8) can draw cards that scales with the number of bulbs on your board (which could be significant).

At the end, a single card is selected from the list of cards and melded.  Currently, 3 cards can be selected and remembered (via card_id_x).  This needs to be enhanced for this card to be implemented (or a different solution needs to be implemented).

ultimatefiend:
it might make sense to have a different location...sort of like "revealed", but you don't reveal.  It would also be considered part of the hand, but would only be visible to the player.  This could also be used for lever as well

micahstairs:
That's a neat idea! Should it behave the same as the revealed zone (except that it isn't actually visible to other players)? If so, will players think we are accidentally revealing their cards?

ultimatefiend:
hmm good point.  the log should be clear that the cards aren't being revealed....or maybe it is more like a subset of the hand?  maybe it appears in a horizontal line above your hand or something like that.

ultimatefiend:
the other player wouldn't need to see what is going on at all.  it could be local to the current player.

Issue #489, opened 2022-06-05
Hot Air Balloon requires the ability to select specific cards on other player's boards


ultimatefiend:
This is the card text:

You may achieve (if eligible) a top card from any player\'s board if they have an achievement of matching value. If you do, transfer your top green card to that player\'s board. Otherwise, draw and meld a ${age_7}.

The first part of being able to select a specific card of a specific value on other player's boards isn't covered currently.  Charter of Liberties lets you select splayed piles on other boards (for example), but doesn't allow the selection of specific ages.

In this case, the ages are limited to the achievement ages that a specific player has (these numbers will be different based on what each player has achieved).

micahstairs:
Hmm maybe I will need to add a special "matches_achievement_value" option.

PR #585, opened 2022-09-13
#489, #503

ultimatefiend:


ultimatefiend:
As noted in the TODO, there is work to be done with the required achievement eligibility flag when the base achievement isn't available anymore.

micahstairs:
I'll merge this in and work on a fix in a separate PR.

Also I noticed that Charitable Trust has the same bug as Hot Air Balloon so I also left a TODO there.

micahstairs (approved):

Issue #355, opened 2022-04-09
Render Echo effects on the cards


ultimatefiend:
Echo effects have different text based on the card.

I think we talked about having a single icon value for the echo effect, but then the specific text would be written in the box based on the card.

"Candles" (Age 1 Red) has the longest string to put in an Echo effect box.

'echo_effect_1' => clienttranslate('If every other player has a higher score than you, draw a ${age_3}.')

PR #591, opened 2022-09-18
Stylize echo effects (#355)

micahstairs:
I opted for a simpler look than what's on the real cards (this leaves more room for the text, and saves us from adding more images to the game, especially since many people dislike the checkerboard pattern to begin with).

It looks good on both the first and third edition card backgrounds:

Screen Shot 2022-09-18 at 12 20 21 PM
Screen Shot 2022-09-18 at 12 20 27 PM

### DIFF
--- a/innovation.css
+++ b/innovation.css
@@ -888,6 +888,31 @@
   height: 151px;
 }
 
+.echo_effect {
+  border-style: solid;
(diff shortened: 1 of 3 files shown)


ultimatefiend (approved):
looks great!

Issue #445, opened 2022-05-11
Transferring a card from hand to 'any other player' causes an error


ultimatefiend:
this:

        case "347N1A":
            // "Transfer a card from your hand to any other player's board."
            $options = array(
                'player_id' => $player_id,
                'n' => 1,
                'can_pass' => false,

                'owner_from' => $player_id,
                'location_from' => 'hand',
                'owner_to' => 'any other player',
                'location_to' => 'board',
            );
            break;

is causing this:

Unexpected error: Error while processing SQL request (Localhost via UNIX socket): UPDATE global SET global_value='any other player' WHERE global_id='35' Incorrect integer value: 'any other player' for column 'global_value' at row 1 (reference: GS1 11/05 06:06:43)

"Crossbow" will cause this error to occur.

micahstairs:
So I looked into this, `any other player` is only supported by `owner_from`, not `owner_to`. Can you break this interaction into two steps? The first part is choosing the opponent and then the second part is choosing the card to give them. Because of automation, the first part would always be skipped by automation anyway.

PR #602, opened 2022-09-20
#417, #445, #501

ultimatefiend:


micahstairs (approved):
Nice work!

I ended up making some fairly big changes, so I retested each of the cards myself.

PR #613, opened 2022-09-25
#501

ultimatefiend:
Note that Laser's draw and meld 10s is done twice because the "B" second step is suppressed when 'n' is zero.


micahstairs (approved):
I only made a couple minor changes!

Issue #360, opened 2022-04-09
Special hex icons need to be rendered in some dogma effects 


ultimatefiend:
Some cards (e.g. Loom, red age 6) show an empty black hex icon in the dogma text.  For the moment, this was specified as ${icon_0} where it is needed.

Also, Saxophone Age 7 has a special hex icon in its dogma text (the hex has a note in the middle).

Please add these to hexagon_icons_echoes.png (you could make them the same size as the other hex icons for simplicity, even though we will shrink them in-game).

micahstairs:
Josh did his part, now I'll take care of getting this incorporated into the front-end.

micahstairs:
We can't forget about the Heritage tooltip too.

PR #620, opened 2022-09-28
add special hexes for saxophone and empty hex

normaljosh:
Adding special hexes for empty spaces and hexagon note icon as described in #360


micahstairs (approved):
Amazing! Thanks for tackling this so fast.

Issue #621, opened 2022-09-28
Jet, Karaoke, Charitable Trust, and Stethoscope behave inconsistently when shared


madisonsilver:
Stethoscope is supposed to make you draw an 8 iff you melded a blue card, but it doesn't behave properly when shared.  This can be seen on move 107/108 of https://studio.boardgamearena.com/table?table=414483 (both players draw an 8, even though only MadisonSilver0 should) and throughout https://studio.boardgamearena.com/gamereview?table=412783.

ultimatefiend:
This is the same as #619.  This same issue will affect Karaoke as well.  Basically, sharing the effect corrupts the variable that stores the card (or color) of the card that was melded in the echo effect (or drawn by Charitable Trust).

ultimatefiend:
I realized that if Jet were shared with someone (in a 3 person game for example), then the melded color would get corrupted by a similar mechanism.

micahstairs:
I'll add a player-based auxiliary value. Right now the auxiliary value is shared between players, which causes the corruption.

PR #624, opened 2022-09-29
Fix Jet and Karaoke (#621)

micahstairs:


ultimatefiend (commented):
should this also factor in what needs to happen when endorsing?

For these cards, we will need to store the results of the execution of both echo effects for the originating player.

We could also ignore this for now and consider it when polishing up cities.

micahstairs:
Good question! I think I'd prefer to defer endorsing for now, yeah. We could easily add another column to the table if needed, or encode the 2nd endorsement into the index_id somehow

ultimatefiend (approved):

Issue #639, opened 2022-10-01
Watermill didn't repeat effect


micahstairs:
Also the card with the bonus needs to be revealed
<img width="675" alt="Screen Shot 2022-09-30 at 9 43 16 PM" src="https://user-images.githubusercontent.com/7231485/193378238-03a21dff-1373-4552-8e7c-b46e427cf0ad.png">

micahstairs:
The reveal shouldn't actually happen. From kahlia:

> If the player decides not to return a card to activate the repeat, why should they reveal the card they drew?
> 
> If the player decides to do it, then the card is revealed when it is tucked so I don't think the reveal is really necessary.

PR #642, opened 2022-10-01
Fix Watermill effect repeating (#639)

micahstairs:


ultimatefiend (commented):
If the player decides not to return a card to activate the repeat, why should they reveal the card they drew?

If the player decides to do it, then the card is revealed when it is tucked so I don't think the reveal is really necessary.

micahstairs:
Great point! I removed the reveal. Please take another look!

ultimatefiend (approved):

Issue #609, opened 2022-09-25
Make sure automation handles new options (e.g. with_bonus) correctly in regards to hidden information


micahstairs:
There are a number of new options that need to be considered in the automation rules.

`card_ids_are_in_auxiliary_array` and `with_bonus` are a non-exhaustive list.

PR #644, opened 2022-10-01
Adjust automation so that new Echoes options do not reveal hidden info (#609)

micahstairs:


ultimatefiend (approved):

ultimatefiend (commented):
Actually, there are slightly different rules for revealing for values with bonuses and not having bonuses because not all cards can have bonuses (i.e. base and artifact cards cannot have bonuse).

Does this change take this into account?

micahstairs:
No it does not take that into account. One of my other open PRs actually has a TODO to revisit that revealing logic involving bonuses. I think I'd like to get Chris' input on that before we change either of these spots. But I do agree that we will want consistency.

micahstairs:
I just sent an email to Chris.

Issue #656, opened 2022-10-04
Email interacts strangely with purple cards


madisonsilver:
Email's second dogma effect will sometimes fail to execute purple cards or execute a different card instead.

https://studio.boardgamearena.com/gamereview?table=412005 (moves 6+)
https://studio.boardgamearena.com/gamereview?table=411994 (moves 189+)

ultimatefiend:
It looks like the nested execution doesn't execute a card (or its echo effects) if it has a demand effect.  I can look more into this later.

ultimatefiend:
From the code :

                // "Execute all non-demand dogma effects on your lowest non-green top card. Do not share them."
                // TODO(ECHOES): There's multiple bugs here. Battleship Yamato's faceup age is 11, which breaks an
                // assumption that this code currently has. Also, purple is being exlcuded in the for loop. I propose
                // rewriting this using the countCardsInLocationKeyedByColor function and using an if statement instead
                // of a for loop.

ultimatefiend:
Interestingly, we don't want to execute Yamato so probably excluding it would work just fine for this card.

micahstairs:
Yamato is also a red card, not green. So yeah, I agree, we don't need to worry about that there.

PR #660, opened 2022-10-04
#656

ultimatefiend:


micahstairs (approved):
Made a few changes. Instead of using a magic constant (11), I used null to make the logic more clear.

There's no need to intentionally avoid Battleship Yamato since executing it is a no-op anyway.

Issue #657, opened 2022-10-04
Candles warns when demand won't do anything even though echo effect will


jacksonmead:
When you dogma Candles and the demand isn't going to fire, you get the warning about the card not having any effect, even if the echo will have an effect. I assume this might happen with all cards like this?

ultimatefiend:
Yes, it does.  See #484

Issue #484, opened 2022-06-01
Piles with echo effects need to suppress the "This card won't do anything" warning when a card is dogma'd


ultimatefiend:
This occurred with some demand effects.

PR #662, opened 2022-10-05
Incorporate echo effects into no-op detection (#484, #657)

micahstairs:


ultimatefiend (changes requested):
Piano won't have an effect if all hands are empty.

Tuning Fork will always have an effect.

ultimatefiend (approved):

Issue #622, opened 2022-09-28
Promotions from Forecast in replays cause JavaScript errors


madisonsilver:
Replaying a game with a promotion from forecast can cause a BGA and JavaScript error on the turn with the promotion (both pasted below).

This can be reproduced on the following games:
https://studio.boardgamearena.com/table?table=412783
https://studio.boardgamearena.com/table?table=414483
(It may be possible to reproduce it on https://studio.boardgamearena.com/table?table=415273, but that game replay isn't working at all for me).

```
Javascript error:
Script error.
Script:
Url: https://studio.boardgamearena.com/archive/replay/220928-1000/?table=412783&player=2365492&comments=2365492;
BGA version 220928-1000
U=2365492
Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/105.0.0.0 Safari/537.36
```

```
Uncaught TypeError: Cannot read properties of null (reading 'style')
    at c (dojo.js:1:52644)
    at i.set (dojo.js:1:53342)
    at e.style (dojo.js:1:70522)
    at Object.show (ly_studio.js:1:461668)
    at Object.show (ly_studio.js:1:322119)
    at Object.onEnteringState (innovationdev.js?1664387927876:983:51)
    at Object.ntf_gameStateChange (ly_studio.js:1:639258)
    at r.<anonymous> (dojo.js:1:23159)
    at r.a [as ongameStateChange] (dojo.js:1:43047)
    at o.emit (dojo.js:1:39041)
```

micahstairs:
I managed to replicate it. The line triggering this is `this.my_forecast_verso_window.show();`.

I'm curious if the same bug happens when a player is prompted to click a card in their personal score pile (since all of the forecast code for the popup was just copied from that).

micahstairs:
Yeah I managed to replicate the same bug with the score dialog:

`this.my_score_verso_window.show();`

PR #668, opened 2022-10-07
Fix replay bug involving score/forecast popups (#622)

micahstairs:


ultimatefiend (approved):
It looks like this should fix vaccination and chemistry as well.

PR #675, opened 2022-10-07
fix watermill, etc

ultimatefiend:
barometer updates
tuning fork first attempt


ultimatefiend:
Both Barometer (2nd dogma effect) and Tuning Fork (echo effect) aren't working.

The rest of it seems to work.  You might have a better way to handle tuning fork's dogma effect.  I replicated the code so the subsequent dogmas could be optional (the first one is not).

micahstairs (approved):

Issue #645, opened 2022-10-01
Remove code branches for release_version = 0


micahstairs:
There's currently only 1 game left which started before July 13th, 2022, so we are on track to remove release_version=0 code branches and eliminate some tech debt!

We will lump this into the Echoes release so that these changes can get some extra playtesting before it hits production.

micahstairs:
Also I just checked and there are no longer any games left which started before July 13th, which means the above changes will be safe to deploy!

PR #684, opened 2022-10-09
Remove unused global variables (#645)

micahstairs:


ultimatefiend (commented):
looks ok.  is 'auxiliary_value' => 49, really deprecated?   Is it stored in the nested table now?

micahstairs:
Correct! We only use the nested table now.

ultimatefiend (approved):

Issue #685, opened 2022-10-09
Make sure all Echoes TODOs are complete


micahstairs:
These are the ones I found:

	Line   163:         // TODDO(ECHOES): Manually test this.
	Line   165:             // TODDO(ECHOES): Make sure newly added global variables get added here.
	Line  2226:             // TODO(ECHOES,FIGURES): Update this if any cards transfer non-relic cards from a player's achievement pile to their hand.
	Line  7991:         // TODO(ECHOES#597,FIGURES): Add cases.
	Line 13096:                         // TODO(ECHOES#472): The value here could be as high as 13 with a visible bonus of 11
	Line 13751:                 // TODO(ECHOES): Instead of clicking buttons to choose the top or bottom card, it would a better user experience
	Line 13838:                     // TODO(ECHOES): Update these log statements to include the music note icon.
	Line 19780:             // TODO(ECHOES): This needs to have the "choose_draw_value" when that is implemented since 11s can appear as bonuses
	Line 20255:             // TOOD(ECHOES): Test that this behaves correctly with Battleship Yamato.
	Line 23965:                                 // TODO(ECHOES): It's not clear whether we should be checking for other matching icons too (e.g. bonus icons).

ultimatefiend:
case 'achievements->hand':
            // TODO(ECHOES,FIGURES): Update this if any cards transfer non-relic cards from a player's achievement pile to their hand.
            $message_for_player = clienttranslate('${You} seize ${<}${age}${>} ${<<}${name}${>>} from your achievements to your hand.');
            $message_for_others = clienttranslate('${player_name} seizes the ${<}${age}${>} relic from his achievements to his hand.');
            break;

There is no way to remove achievements in Echoes.

PR #702, opened 2022-10-13
#685

ultimatefiend:
Fix one of the todos (the "must") for sunglasses.

Also fix a corner case where no non-purple piles are splayable (i.e. have one card) and thus cannot match a purple splay.


micahstairs (approved):
I might revisit this implementation later to see if I can simplify it.

Issue #422, opened 2022-04-29
Revise calculations for placement of score and forecast zones


micahstairs:
When I added the forecast zone to the UI, I didn't spend any time getting the calculations right. Ideally there would be one card's worth of space between the two zones.

We also need to make sure that when Echoes is disabled that the forecast zone is removed from the board and the calculations still make sense.

We also need to make sure the layout looks good on both desktop and mobile.

ultimatefiend:
it sits a little low on the player board.  Was that intentional:

![image](https://user-images.githubusercontent.com/11416698/166085930-6d29fe30-929e-4e28-8029-3f5d506bc1c6.png)

micahstairs:
You need to force refresh your page.

ultimatefiend:
Ah yes, that helped.  There is an offset with the scorepile though:

![image](https://user-images.githubusercontent.com/11416698/166121073-90ecf824-0175-4d0e-ace6-c63419094197.png)

micahstairs:
Yeah I saw the same offset but couldn't quickly figure out a fix. I was focusing my time on just getting the zones functional haha.

micahstairs:
Here's a screenshot to capture the weird overlap that's happening:

<img width="826" alt="Screen Shot 2022-10-23 at 8 21 07 PM" src="https://user-images.githubusercontent.com/7231485/197425666-3769ac0c-270d-4ecc-b06e-8b5abc3b7133.png">

micahstairs:
Before closing this, I need to shrink the achievement zone to a width of 2 cards on mobile when the forecast is present.

micahstairs:
![Screenshot_20221207-171943](https://user-images.githubusercontent.com/7231485/206309120-538922ac-40db-4672-bb08-2da1d9f46762.png)

micahstairs:
For some reason the score and forecast zones keep alternating between 2 and 1 cards. This is the last thing we need to fix.

micahstairs:
@lordmcfuzz I'm taking this back! I see the problem and will fix it.

PR #730, opened 2022-10-24
Fix score and forecast zone layout (#422)

micahstairs:
Before:
Screen Shot 2022-10-23 at 8 21 07 PM

After:
Screen Shot 2022-10-23 at 8 46 51 PM

### DIFF
--- a/innovation.js
+++ b/innovation.js
@@ -358,9 +358,6 @@ function (dojo, declare) {
                 this.num_cards_in_row.score = this.num_cards_in_row.score / 2;
             }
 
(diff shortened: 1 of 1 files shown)


ultimatefiend (approved):
Does this work on mobile as well?

ultimatefiend:
let me know when you push this to innovationdev.  I will see how it looks on my phone.

micahstairs:
I'm not sure how it looks on mobile yet, but okay, we can push it and iterate if there are problems!

PR #738, opened 2022-10-31
Example of using bga-wb and inversion of control

splittingred:
## What?

* Adds the ability to test against the database. Detailed in README, but copy `bgaproject.dist.yml` to `bgaproject.yml`, set your settings, then run `./vendor/bin/phpunit` to see it in action.
* Note that the bgaproject.yml implementation here is just for testing; not used in actual runtime (could be! it has deploy functionality but not using it now)
* Adds a new `Card` model and `\Innovation\Cards` service for getting a Card model, which can then be used elsewhere (with programmatic type enforcement!)
* Added a `getDatabaseConnection` method to the main `\Innovation` class to allow to inject that into other classes (such as a service class) so you can issue DB queries without needing the main class to do so. (BGA uses Doctrine's DBAL under the hood)
* Using my fork of the bga-workbench which has a bunch of fixes and quality-of-life improvements to it, which also allows it to work on Github CI (for automated tests)

I don't have a separate Innovation game on BGA setup at present, so I didn't test actually trying this out on a deployed Innovation game instance on BGA. But, there was some discussion in the Discord about this so I thought I'd show how it can happen.

### DIFF
--- a/innovation.game.php
+++ b/innovation.game.php
@@ -17,10 +17,11 @@
   */
 
 
+use Doctrine\DBAL\Connection;
+use Innovation\Cards;
+
 require_once(APP_GAMEMODULE_PATH.'module/table/table.game.php');
-require_once('modules/Innovation/Utils/Arrays.php');
-require_once('modules/Innovation/Utils/Strings.php');
-require_once('modules/Innovation/GameState.php');
+require_once('modules/loader.php');
 
 /* Exception to be called when the game must end */
 class EndOfGame extends Exception {}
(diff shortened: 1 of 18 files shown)


innovation.game.php
```diff
@@ -6093,7 +6114,7 @@ function initialMeld($card_id) {
         }
         
         // Mark it as selected
-        self::markAsSelected($card_id, $player_id);
+        self::markAsSelected($card_id);
```
splittingred, line 6117: 💁 This was removed as there is no actual 2nd argument to `markAsSelected`

micahstairs:
Note to self: this was added to the wrong branch (main-dev) l. I'll need to try cherry picking this onto main-echoes.

micahstairs:
Thanks for the getting the ball rolling on this! This didn't deploy successfully. Please see https://github.com/micahstairs/bga-innovation/tree/micahstairs-splittingred (I fixed some of the errors, but there are still some, as I mentioned to you over Discord). I'll let you dig further into it

Issue #777, opened 2022-11-29
Minimize how often we set enable_autoselection to false


micahstairs:
I recently added enable_autoselection=false to a bunch of cards in light of the recent special achievement check changes. However, I need to revisit this since we should still be able to autoselect if there's only one possible option (we just don't want to meld or return cards if the selection size is at least 2).

PR #779, opened 2022-11-29
Automate single card returns and single card melds (#777)

micahstairs:
I recently added enable_autoselection=false to a bunch of cards in light of the recent special achievement check changes. However, I had to revisit that since we should still be able to autoselect if there's only one possible option (we just don't want to meld or return cards if the selection size is at least 2).


micahstairs:
I just tested and this fixes https://boardgamearena.com/bug?id=76650 too.

ultimatefiend (approved):

PR #788, opened 2022-11-30
Fix bug preventing player from getting winning special achievement while promoting a card

micahstairs:


micahstairs:
I'm going to self-approve this because I have an ongoing alpha game where I'm encountering this error (preventing me from winning).

PR #790, opened 2022-12-02
Show preview of score after splaying (https://boardgamearena.com/bug?id=76738)

micahstairs:
Screen Shot 2022-12-01 at 9 03 08 PM


ultimatefiend (commented):
should this happen when echoes and/or cities are not present?  It looks like it will.  I like the idea. :)

micahstairs:
Oh good call! I updated it so that the new message only appears if Cities or Echoes are in use.

ultimatefiend (approved):

PR #796, opened 2022-12-04
Make it more obvious when an Echoes card is going to be drawn

micahstairs:
This was requested in //boardgamearena.com/bug?id=76960.

Screen Shot 2022-12-03 at 9 11 29 PM
Screen Shot 2022-12-03 at 9 11 39 PM


ultimatefiend (commented):
will this be translated?  If so, looks good.

micahstairs:
Yup! The `_()` part triggers translation.

ultimatefiend (approved):

PR #805, opened 2022-12-08
Fix bug in Marcha Real and add more logging

micahstairs:
Reported in https://boardgamearena.com/bug?id=75931.

### DIFF
--- a/innovation.game.php
+++ b/innovation.game.php
@@ -11872,7 +11872,7 @@ function getOwnersOfTopCardWithColorAndAge($color, $age) {
             // id 174, Artifacts age 6: Marcha Real
             case "174N1":
                 self::setAuxiliaryValue(-1);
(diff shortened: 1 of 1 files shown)


innovation.game.php
```diff
@@ -11871,6 +11871,8 @@ function stPlayerInvolvedTurn() {
                 
             // id 174, Artifacts age 6: Marcha Real
             case "174N1":
+                self::setAuxiliaryValue(-1);
+                self::setAuxiliaryValue(-1);
```
ultimatefiend, line 11875: why is this done twice?

micahstairs: Thanks for catching this bug! I meant to use setAuxiliaryValue2 the second time.

ultimatefiend (approved):

Issue #533, opened 2022-07-24
Special achievement text not visible on mobile


ultimatefiend:
"Mouse over" doesn't really exist on mobile so it is difficult to tell what special achievements are available or in a player's achievement pile.

This was mentioned in a forum post (but not a bug report).

A good solution would be to imitate what other games do when trying provide details on cards like in Agricola:

![image](https://user-images.githubusercontent.com/11416698/180652500-952a8552-5347-4c3e-a098-609bf47cd16d.png)

There is a button to provide details on the card.  This would work with the specials because nothing happens when you click on a card in normal situations.

micahstairs:
You can actually press and hold to make it appear, but I agree there's room for improvement here.

micahstairs:
For the record, this was later filed as a bug report (https://boardgamearena.com/bug?id=68916).

PR #807, opened 2022-12-09
Add feature which allows players to browse list of special achievements (#533)

micahstairs:
This has been a big request from players! https://boardgamearena.com/bug?id=68916

Screen Shot 2022-12-08 at 8 06 13 PM

### DIFF
--- a/innovation.css
+++ b/innovation.css
@@ -117,7 +117,7 @@
   height: 49px;
 }
 
-#change_display_mode_button {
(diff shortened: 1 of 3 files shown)


ultimatefiend:
I am curious what this will look like on mobile.  I will give it a try when its live.

micahstairs:
I was also curious! Alpha is the perfect place to test that.

Issue #791, opened 2022-12-02
Show preview of new score on meld tooltip


micahstairs:
In https://github.com/micahstairs/bga-innovation/pull/790, I did this on the splay tooltip, but we want a similar thing on the meld tooltip.

PR #811, opened 2022-12-09
Add score calculation to meld tooltip and fix bug in splay tooltip (#791)

micahstairs:
Screen Shot 2022-12-09 at 3 20 05 PM

The splay tooltip wasn't looking for bonuses underneath the other piles (the ones not about to be splayed).

Also, while I was here, I knocked out a few TODOs for Cities.

### DIFF
--- a/innovation.js
+++ b/innovation.js
@@ -1798,6 +1798,23 @@ function (dojo, declare) {
         },
         
         createActionTextForMeld : function(self, card) {
+            // Calculate new score (score pile + bonus icons)
+            var bonus_icons = [];
+            for (var i = 0; i < 5; i++) {
+                var pile_zone = self.zone.board[self.player_id][i];
+                var splay_direction = pile_zone.splay_direction;
+                // If there are no cards in the stack yet, treat the pile as unsplayed
+                if (pile_zone.items.length == 0) {
+                    splay_direction = 0;
+                }
+                if (i == card.color) {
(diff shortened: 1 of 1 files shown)


innovation.js
```diff
@@ -1806,61 +1823,80 @@ function (dojo, declare) {
...
+                        {
+                            'age': self.square('N', 'age', top_card.age, 'in_log'),
+                            'card_name': "<span class='card_name'>" + _(top_card.name) + "</span>"
+                        });
+                }
+            } else {
+                if (self.cities_expansion_enabled || self.echoes_expansion_enabled) {
+                    HTML_action += "<p>" + dojo.string.substitute(_("If you do, you will have a total score of ${score} and your new featured icon counts will be:"), {'score' : new_score}) + "</p>";
+                } else {
+                    HTML_action += "<p>" + _("If you do, your new ressource counts will be:") + "</p>";
```
ultimatefiend, line 1844: spell check here

micahstairs: That's an existing string that is already translated so I couldn't fix the typo without triggering retranslation.

ultimatefiend (approved):
I think it looks good.  Would it make sense to also show the number of hexes and the number of bonus icons as well (for the special achievements)?

micahstairs:
Can you file an issue to track that as a future improvement? I think the score is most important for now, and I don't think I'll get to it before release.

PR #814, opened 2022-12-10
Show current progression towards each special achievement

micahstairs:
Screen Shot 2022-12-09 at 10 30 37 PM

### DIFF
--- a/innovation.css
+++ b/innovation.css
@@ -140,6 +140,11 @@
   max-width: 45px;
 }
 
+.special_achievement_status {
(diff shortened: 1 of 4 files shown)


ultimatefiend (approved):
Could the front end ask the back end for some of this information?  It looks like JS is redo-ing the logic for detecting the special achievements.  I guess it doesn't hurt too much to do it twice.

micahstairs:
That would unfortunately be a lot more complicated.

PR #816, opened 2022-12-10
Add small 'Browse' button near special achievements

micahstairs:
Screen Shot 2022-12-10 at 11 38 06 AM

I also noticed that the special achievements were slightly off-center, so I fixed that at the same time.

### DIFF
--- a/innovation.css
+++ b/innovation.css
@@ -125,6 +125,14 @@
   border-style: dotted;
 }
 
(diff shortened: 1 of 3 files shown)


micahstairs:
We may want to remove the other button altogether and just keep this one?

ultimatefiend (approved):

PR #824, opened 2022-12-12
Remove TODO (#685)

micahstairs:
I checked and cards only end up in the forecast via a "draw and foreshadow" mechanism, so it's not possible for Artifacts to end up in the forecast anyway.


micahstairs:
Haha yeah Figures is going to mess up all of our assumptions anyway!

PR #837, opened 2022-12-18
Add card revealing to Typewriter (https://boardgamearena.com/bug?id=77541)

micahstairs:


ultimatefiend (approved):
nitpicking here...but shouldn't only the unique colored cards be revealed for the purposes of drawing cards?

this is fine for now, but I thought i'd bring it up.

micahstairs:
I see your point, but then we'd need to make the card way more complicated in order to let the player choose which cards to reveal.

PR #840, opened 2022-12-19
Add debug buttons to splay piles

micahstairs:
I should have implemented this a long, long time ago...

Screen Shot 2022-12-18 at 8 14 53 PM


ultimatefiend (approved):
nice addition!  I was using publications to get splays before.

Issue #646, opened 2022-10-01
Adjust automation logic in Oars when Echoes drawing rules make it so that the order makes a difference


micahstairs:
When Echoes is in play, there are certain situations where Oars can't simply transfer the cards in an arbitrary order. We need to reduce the automation in those situations.

ultimatefiend:
Are there other cards where the order of drawing would be affected?

micahstairs:
Off the top of my head, I don't think so. Oars has some custom automation for this special case (a series of transfers and draws). The general automation engine won't cause any problems.

PR #850, opened 2022-12-23
Fix automation bug in Oars (#646)

micahstairs:
Also added a new `usingThirdEditionRules` function to improve code clarity.


ultimatefiend (approved):
looks good. the utility will be helpful in the future I think. :)

PR #853, opened 2022-12-24
Temporarily disable Cities option in preparation for Echoes release

micahstairs:


ultimatefiend (approved):
do you really need to remove all of the cities implementation if it is basically hard coded to disabled in the settings?  It seems like that adds unnecessary work.

Not blocking, but I'm curious.

micahstairs:
Removing it for the release seems cleaner. Otherwise players may get confused why they can't enable it.

ultimatefiend:
I understand that part.  You can just disable it in the code at startup and leave everything else alone.

micahstairs:
That was my initial plan! But other parts of the code depend on that `cities_mode` global variable existing.

This will be an easy revert later!

micahstairs:
Oh I see what you're suggesting now. I'm not sure if it will let me keep variable 103 defined without also having the option there, but I will try! If it works, then this becomes much simpler.

micahstairs:
It worked! I merged it in here: https://github.com/micahstairs/bga-innovation/commit/e93ede088fe2ee36b1e41908ca4bb1164412e9d0.

PR #859, opened 2022-12-26
Render special icons on Cities special achievements (#384)

### LINKED ISSUE #384: Implement tooltip for Cities special achievement cards
We need the proper icons for the achievements (the arrows, flags, and fountains).

There also isn't a method for claiming these in an alternative way which is different from special achievements in Base and Echoes.

micahstairs:
Screen Shot 2022-12-26 at 5 06 34 PM
Screen Shot 2022-12-26 at 5 08 35 PM

Instead of adding in new brown images like we see on the real cards, I just decided to use a grayscale version of one of the colored icons. It does the trick!

This closes #384.

### DIFF
--- a/innovation.css
+++ b/innovation.css
@@ -1470,7 +1470,7 @@
 .square.in_tooltip.icon_8 {
   background-position-y: -42px;
 }
-.square.in_tooltip.icon_8 {
+.square.in_tooltip.icon_9 {
   background-position-y: -63px;
 }
 .square.in_tooltip.icon_11 {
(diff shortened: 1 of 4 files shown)


ultimatefiend (changes requested):
Victory should be a flag and not a plus, but otherwise looks good.

micahstairs:
Screen Shot 2022-12-27 at 12 55 47 AM

Thanks for catching that! I had a typo in the SCSS/CSS.

ultimatefiend (approved):

Issue #632, opened 2022-09-30
Jet, Karaoke, Charitable Trust, and Stethoscope need to be tested with endorse feature


ultimatefiend:
The main dogma effect is reliant upon the results of the card's echo effect so the endorse feature will need to be specially handled by these cards.

ultimatefiend:
add Dentures to the list.

Watno said that the main effect should offer a choice for some of these effects (e.g. Karaoke) when the endorse is used.

micahstairs:
I did a full audit of cards we should check:
Oars, Mapmaking, Gunpowder, The Pirate Code, Vaccination, Democracy, Dancing Girl, Charitable trust, Kobukson, Dentures, Morphine, Stethoscope, Jet, Karaoke

We should also make sure these cards behave correctly when the card has been executed twice in the current nesting stack (e.g. Democracy is not behaving correctly in this regard).

micahstairs:
I found a few relevant BGG threads:
- https://boardgamegeek.com/thread/1093137/endorsing-democracy
- https://boardgamegeek.com/thread/2721417/endorse-connected-actions

madisonsilver:
Endorsing Charitable Trust gives the following error on the second iteration of the echo effect:

`Unexpected error: Error while processing SQL request (Localhost via UNIX socket): INSERT INTO indexed_auxiliary_value (nesting_index, index_id, value) VALUES (0, 2365493, 42) Duplicate entry '0-2365493' for key 'PRIMARY' (reference: GS1 03/01 00:31:33)`

This makes it impossible to proceed with the game (see e.g https://studio.boardgamearena.com/table?table=439017 )

ultimatefiend:
Karaoke just triggers the second card that is melded.  It doesn't execute the first card that was melded.

You have to execute the 2nd non-demand effect.
Execution of 8 Rocketry is complete.
No card matches the criteria of the effect.
You have 7 .
You have to execute the non-demand effect.
You execute the non-demand effect(s) of 8 Rocketry.
You have to execute the 1st non-demand effect again because it was endorsed.
Execution of 8 Rocketry is complete.
No card matches the criteria of the effect.
You have 7 .
You have to execute the non-demand effect.
You execute the non-demand effect(s) of 8 Rocketry.
You have to execute the 1st non-demand effect.
You draw and meld 8 Rocketry.
You choose the value 8.
You have to execute the echo effect again because it was endorsed.
You draw and meld 9 Fission.
You choose the value 9.
[...]

PR #886, opened 2023-01-10
Fix bug Kobukson when the demand executes more than once (#632)

micahstairs:
This bug would happen if the demand was requested of multiple players, or if the dogma was endorsed.


ultimatefiend (approved):
could Jet be done similar to this?  This seems pretty simple.

PR #887, opened 2023-01-10
Fix endorsement of Jet (#632)

micahstairs:
I manually tested a bunch of cases:

Screen Shot 2023-01-10 at 2 28 14 PM
Screen Shot 2023-01-10 at 2 28 56 PM
Screen Shot 2023-01-10 at 2 34 02 PM

### DIFF
--- a/innovation.game.php
+++ b/innovation.game.php
@@ -14484,19 +14484,33 @@ function getOwnersOfTopCardWithColorAndAge($color, $age) {
                 
             // id 418, Echoes age 9: Jet
             case "418E1":
-                if ($player_id == $launcher_id) {
-                    self::setAuxiliaryValue(-1);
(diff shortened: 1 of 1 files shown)


ultimatefiend (commented):
couldn't the color that is selected be added to a list? (and then the interaction looks at the list)

if its not endorsed, then there would be only color in the list (and no choice needs to be made)

The clearing of the color indexing could be avoided in the event an endorse is happening with the card, but little else would need to be done.

micahstairs:
You're actually describing what I implemented!

We store a list of the colors melded by the launcher, and then for each player, they need their own auxiliary value to indicate which value to remove from the array for their second choice.

ultimatefiend (approved):

PR #903, opened 2023-01-18
Make it obvious in the player panel how many achievements are required to win

micahstairs:
Screenshot 2023-01-18 at 1 42 04 AM

The only wrinkle was that I had to hide the "/X" part if the game ends by score or dogma. Otherwise this would have been a 1-line change.


ultimatefiend (approved):
I think the text below the "achievements" text on the tableau that states how many are required to win could be removed.  It displays poorly on my mobile (it gets cut off) and otherwise just adds clutter.

Otherwise, it is a valuable addition.  I find myself wondering how many more achievements are needed (on mobile) so this will make it crystal clear.

micahstairs:
Good idea! Done.

PR #909, opened 2023-01-21
Indicate when a Meld action will cause a City card to be drawn

micahstairs:
I also fixed a bug where if we tried to draw a City and there weren't any cards left in the deck, we no longer get an error but gracefully fall back to the base deck (or Echoes deck when applicable).

Screenshot 2023-01-21 at 8 20 56 AM

### DIFF
--- a/innovation.game.php
+++ b/innovation.game.php
@@ -5599,6 +5599,12 @@ function getOwnersOfTopCardWithColorAndAge($color, $age) {
             throw new EndOfGame();
         }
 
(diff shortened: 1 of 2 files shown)


ultimatefiend (approved):
this seems to work for empty piles, but what about the cities with arrows on them?

micahstairs:
Oh good point. Just added logic to handle that case!

ultimatefiend (approved):

PR #918, opened 2023-01-29
Use instantaneousMode to make transition more graceful when resizing screen

micahstairs:


micahstairs:
Never mind, this didn't work as intended.

PR #943, opened 2023-02-10
Open forecast popup when player is prompted to select forecast cards

micahstairs:
The forecast popup is now opened automatically if a player is being prompted to select cards from their forecast (https://boardgamearena.com/bug?id=80776).

This mirrors what is done for the score pile already.


ultimatefiend (approved):
I like how it catches replays as well.

PR #1187, opened 2023-07-13
red envelope, brethren

ultimatefiend:


ultimatefiend:
I reread the card and I think the change to brethren might be unnecessary
It really depends on whether the array is initialized or not.

micahstairs:
Actually that Brethren change would actually introduce a bug.

The array is being initialized at the start of the action and we shouldn't reinitialize it.

PR #1213, opened 2023-07-19
fix surveillance

ultimatefiend:


ultimatefiend:
https://boardgamearena.com/bug?id=93608
https://boardgamearena.com/bug?id=93735
https://boardgamearena.com/bug?id=92714

Issue #1251, opened 2023-07-29
Slot machine should use an interaction to return cards


ultimatefiend:
https://boardgamearena.com/bug?id=94469

When the piles are empty, there can be repeat ages so the return order can be important.

Issue #1250, opened 2023-07-29
Ninja claimed the wrong player chose a color


ultimatefiend:
I suspect that this is related to notification for the demander getting confused with the person being demanded in the transfer logic.

micahstairs:
Link to bug: https://boardgamearena.com/bug?id=93980

ultimatefiend:
This is confirmed.

ultimatefiend:
It looks like the names are reversed for the color selection. (the opposing players show the wrong player making the choice.)

ultimatefiend:
It might be related to the effect being a demand effect and the demander is making the choice and not the main player

ultimatefiend:
Yes, this is incorrect:

  public function handleSpecialChoice(int $color): void
  {
    $this->notifications->notifyColorChoice($color, self::getPlayerId());
    self::setAuxiliaryValue($color);
  }

Issue #1249, opened 2023-07-29
Witch Trial should reveal hand if no colors match


ultimatefiend:
If the no card is returned from hand, the hand should be revealed.
https://boardgamearena.com/bug?id=93841

PR #1252, opened 2023-07-29
fix slot machine (94469)

ultimatefiend:


ultimatefiend:
#1251

ultimatefiend:
#1249 and #1250

micahstairs:
Thanks for these fixes!

PR #1258, opened 2023-08-01
#93956: "Jackalope did not allow me to choose the color to unsplay"

Iain27:
Changed the getColorsWithMostVisibleCards function to check if card are splayed otherwise during sharing the player is prompted to unsplay colors not splayed when there are two pile the same. Secondly if more than 1 color has the same number of visible cards add to the array instead of replacing the array.


Iain27:
https://boardgamearena.com/bug?id=93956

micahstairs:
Thanks! I tweaked it a bit since no colors will ever have more cards than `mostVisibleCards`, since we already looped through the array to compute that.

Issue #413, opened 2022-04-26
Use .jpg file for hex icons instead of .png


micahstairs:
Right now we are using a .png file for the hex icons so that they background can be transparent, however, this makes the file very large. We should try changing this to .jpg and using clipping paths instead of relying on the transparent background.

Something like:
```
clip-path: polygon(50% 0%, 100% 25%, 100% 75%, 50% 100%, 0% 75%, 0% 25%);
```

micahstairs:
We should cross reference the support with the list of BGA supported browsers. See https://boardgamearena.com/faq.

micahstairs:
From skimming through the bga-developers discord, I don't think we need to worry about IE. So I'll mark as unblocked.

PR #1357, opened 2023-09-23
Crossword(#413) Bug#99413

massir-rpi:
Crossword would execute the 'if foreseen' effect even on regular dogmas. This was solved by adding a missing 'ifForeseen()' conditional.


micahstairs (approved):
Thanks!

PR #1366, opened 2023-09-28
Map Locations::AVAILABLE_ACHIEVEMENTS to Locations::ACHIEVEMENTS

massir-rpi:
Several cards (366, 385 482) were bugged due to failed getCards() calls that referenced a Locations::AVAILABLE_ACHIEVEMENTS const that did not get mapped back to Locations::ACHIEVEMENTS (owner = 0). In the getOrCountCardsInLocation() function, this mapping is now being done, which fixes these bugs.


micahstairs:
Thanks!

PR #1445, opened 2023-12-19
Update AbstractCard so base splay override returns a bool.

Arctem:
This should fix Touchscreen thinking that no cards have been splayed: https://boardgamearena.com/bug?id=104361


micahstairs:
Nice find!

Issue #1287, opened 2023-08-19
Replays break when cards are junked


micahstairs:

ultimatefiend:
This bug makes it hard to triage issues that are reported.

micahstairs:
Yeah agreed. I really need to get this fixed soon.

micahstairs:
This might possibly be fixed by doing https://github.com/micahstairs/bga-innovation/issues/1268

micahstairs:
I suspect this has to do with the issue where dialogs don't work in replays. In the case of the score and forecast zones, we should don't put the cards on those dialogs. We need to do something similar for the junk zone, but the cards still need to live somewhere. Maybe we need to figure out the underlying issue with the dialogs.

micahstairs:
It's also worth noting that this bug happens when anything is junked, not just decks.

brianwk:
@micahstairs Root cause identified. It's actually not related to junking at all, but rather to bulk transfers and how the event of a bulk transfer bubbles to the frontend. Essentially, another event was created called cardTransferedNoDelay. When I removed the calls to this event and just used cardTransfered, the issue resolved. 

We should create another issue in GitHub for speeding up the transfer of cards. I believe we can simply combine these into one message that lists multiple cards, but I'm also not sure if this message is being generated by us or BGA's SDK.

It sounds like #1268 was what I was describing in my comment here?

PR to revert the change: https://github.com/micahstairs/bga-innovation/pull/1470

brianwk:
@micahstairs I tried a few different things, however without a major overhaul of how the zone mutations occur there will continue to be bugs when multiple cards are moved within one move. The delay on notif_cardTransfered prevents this condition from occurring. It is possible we can lower the delay for now, but I wasn't able to come up with a "quick fix" and this is going to require a major refactor. 

[...]

PR #1470, opened 2025-08-29
revert change to call transferedCardNoDelay event when multiple cards

brianwk:
See my comment in #1287

### DIFF
--- a/src/innovation.ts
+++ b/src/innovation.ts
@@ -5181,6 +5181,7 @@ class Innovation extends BgaGame {
         dojo.subscribe('transferedCard', this, "notif_transferedCard");
         this.notifqueue.setSynchronous('transferedCard', reasonnable_delay);   // Wait X milliseconds after executing the transferedCard handler
         dojo.subscribe('transferedCardNoDelay', this, "notif_transferedCard");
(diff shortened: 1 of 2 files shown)


lordmcfuzz (approved):
Lgtm
Aka the code change looks inoffensive and I didn't test it

micahstairs (changes requested):
Thanks for digging into the (hardest) bug! What happens if we set the "no delay" notifications to 1 instead of 1000? Would that solve the problem without introducing the other issue?

src/innovation.ts
```diff
@@ -1,13 +1,13 @@
 /**
  *------
  * BGA framework: © Gregory Isabelli <gisabelli@boardgamearena.com> & Emmanuel Colin <ecolin@boardgamearena.com>
- * Innovation implementation : © Jean Portemer <jportemer@gmail.com>
+ * innovationbrianwk implementation : © Jean Portemer <jportemer@gmail.com>
```
micahstairs: Make sure to change back innovationbrianwk to innovation

micahstairs (changes requested):

brianwk:
@micahstairs removed the changes to innovation.ts, the changes for now are only in innovation.game.php.

micahstairs (approved):

PR #1471, opened 2025-08-31
#102497: "Generated notifications are larger than 128k"

brianwk:
I trimmed the size of the notifications for each card transfer by 50% it seems. I am not sure if this is enough, but it's a start to improve playability. I didn't see an issue created for this so I made a PR. We may want to test using innovationbrianwk however I tested on BGA Studio and there doesn't seem to be any errors occurring for a wide range of cards. We may be able to remove & infer the player_id / opponent_id from other properties like owner_to / owner_from, but haven't wrapped my head around this yet. We can also shorten some property names, for example changing properties like "owner_from" to "from"

### DIFF
--- a/innovation.game.php
+++ b/innovation.game.php
@@ -59,6 +59,24 @@ class Innovation extends Table
     const COMPEL_EFFECT = 2;
     const ECHO_EFFECT = 3;
 
+    static function stripTransferInfoForNotification(array $transferInfo): array {
+        // dereference transferInfo
+        $transferInfo = (array) $transferInfo;
+        unset($transferInfo['splay_direction_from']);
+        unset($transferInfo['bottom_from']);
+        unset($transferInfo['bottom_to']);
+        unset($transferInfo['score_keyword']);
+        unset($transferInfo['meld_keyword']);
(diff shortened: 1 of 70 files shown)


brianwk:
@micahstairs should we just merge this and see how it goes?

brianwk:
@micahstairs seems like this might be enough.... you can see in this replay I was able to execute Fission and junk over 180 cards simultaneously: https://studio.boardgamearena.com/gamereview?table=766305

brianwk:
@micahstairs One thing I noticed is that the icon are being counted, but the icon isn't being shown in my innovationbrianwk game:
brianwk1 has three .
You have zero .
(It shows exactly like that) -- As far as I know, I didn't remove anything that has to do with the counters (which is stored in progressInfo variable in PHP backend) -- is this just an issue with the renamed files?

Screenshot 2025-08-31 at 11 22 33 PM

This is in the REPLAY log, seen here: https://studio.boardgamearena.com/gamereview?table=766305
Also in the game the Fission execution breaks the replay, but it works in the actual game.

micahstairs (approved):

